PR #1, opened 2023-12-12
Add 'THIS FILE IS AUTOMATICALLY GENERATED. DO NOT EDIT.' comments

inteon:
The "repository-base" module copies a bunch of files in the root of the repo that uses this module.
It also has a verify rule that makes sure these files are not altered.
If anyone wants to make a change to any of these files, it has to be done here in this repo.
This PR adds comments that warn for this.


inteon:
Self-approving this PR, since this is not yet in-use.
/approve
/lgtm

PR #2, opened 2023-12-12
klone has been moved to the cert-manager repo

inteon:
Update the url of the klone tool.


inteon:
Self-approving this PR, since this is not yet in-use.

PR #4, opened 2023-12-18
Add dependabot config to the repository-base module

inteon:
This makes sure that all projects using the repository-base module have dependabot setup correctly.


wallrj (approved):
/lgtm
/approve

modules/repository-base/base/.github/dependabot.yaml
wallrj, line 1: Do we need dependabot to ignore the versions of GitHub actions and Go modules in the `make/_shared/` directory. 
Otherwise we end up with PRs like this:
 * https://github.com/cert-manager/approver-policy/pull/318

What would happen if we merged that PR in cert-manager/approver-policy and merged the same change in this repo.
Would the [make-self-upgrade](https://github.com/cert-manager/makefile-modules/blob/main/modules/repository-base/base/.github/workflows/make-self-upgrade.yaml) workflow fail?

inteon: I think you wouldn't be able to merge that PR because it would cause a validate target to fail.
Ideally, we indeed exclude that folder, but it might not be possible https://stackoverflow.com/questions/65275433/can-i-exclude-directories-from-github-dependabot.
The makefile-modules stuff is quite old and requires some upgrades, also we can still remove some of these go tools if it makes life too hard.

wallrj: Ah, that's a pain. Thanks for looking into it.

PR #5, opened 2023-12-18
Add dependabot for upgrading all own dependencies

inteon:
Adds a dependabot configuration to keep all the dependencies in [makefile-modules](https://github.com/cert-manager/makefile-modules) up-to-date.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,17 @@
+# Update Go dependencies and GitHub Actions dependencies weekly.
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: weekly
+  groups:
+    all:
+      patterns: ["*"]
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+  groups:
+    all:
+      patterns: ["*"]


.github/dependabot.yaml
```diff
@@ -0,0 +1,20 @@
+# THIS FILE IS AUTOMATICALLY GENERATED. DO NOT EDIT.
+# Edit https://github.com/cert-manager/makefile-modules/blob/main/modules/repository-base/base/.github/dependabot.yaml instead.
```
wallrj, line 2: Was **this** file automatically generated? If so, how?
Please document the process in the README file.

wallrj (changes requested):

.github/dependabot.yaml
inteon: No, this file is not auto-generate, but it will be copied into the project & that copy will be auto-generated.

inteon: Indeed, my mistake. @wallrj

wallrj: This file is for updating the dependencies in *this* project, right? e.g.
* https://github.com/cert-manager/makefile-modules/blob/5f23abc93e192e49875d3e7de88ddff00c7331a8/modules/oci-image/image_tool/go.mod#L5-L8

So the comment in this file that says "THIS FILE IS AUTOMATICALLY GENERATED. DO NOT EDIT" is wrong and should be removed.

wallrj (approved):
/lgtm
/approve

PR #6, opened 2023-12-18
Upgrade tools and explain how to upgrade tools

inteon:


.gitignore
```diff
@@ -0,0 +1 @@
+_temp
```
wallrj, line 1: Why `_temp` here rather than `_bin` which is used in other projects?

README.md
```diff
@@ -42,3 +42,10 @@ The order in which the makefiles are includes is as follows (see [Makefile](./mo
 -include make/02_mod.mk
 -include make/_shared/*/02_mod.mk
 ```
+
+### Upgrading the tools in the tools module
+
+1. bump the versions in the modules/tools/00_mod.mk file
+2. run `make tools-learn-sha`
+
+See [Makefile](./Makefile) for more details.
```
wallrj, line 51: Thanks for adding this documentation.

modules/tools/00_mod.mk
```diff
@@ -37,61 +37,61 @@ CTR=docker
...
+TOOLS += cosign=v2.2.2
 # https://pkg.go.dev/github.com/cert-manager/boilersuite?tab=versions
 TOOLS += boilersuite=v0.1.0
 # https://pkg.go.dev/github.com/princjef/gomarkdoc/cmd/gomarkdoc?tab=versions
 TOOLS += gomarkdoc=v1.1.0
 # https://pkg.go.dev/oras.land/oras/cmd/oras?tab=versions
 TOOLS += oras=v1.1.0
 # https://pkg.go.dev/github.com/onsi/ginkgo/v2/ginkgo?tab=versions
-TOOLS += ginkgo=$(shell awk '/ginkgo\/v2/ {print $$2}' go.mod)
+TOOLS += ginkgo=$(shell [[ -f go.mod ]] && awk '/ginkgo\/v2/ {print $$2}' go.mod || echo "v2.13.2")
```
wallrj, line 86: Add an explanatory comment describing:
* **why** we use the version of ginkgo in the `go.mod`?
* What are the circumstances that a `go.mod` file might not be present?
* What happens if the target project has a newer version of ginkgo (perhaps dependabot updated it) in its `go.mod` than `v2.13.2`, in which case the sha sums won't match?

wallrj (changes requested):

modules/tools/00_mod.mk
inteon: 1. because the imported ginkgo version should always match the imported ginkgo version
2. only when we run tools-sha-learn in this repo
3. the ginkgo tool actually does not have a hash defined in this file, we rely on go's cache to make sure that the source was not altered

.gitignore
inteon: For no reason other than to prove that this works too.

modules/tools/00_mod.mk
wallrj: If `ginkgo` the CLI *must* match the version of ginkgo imported by tests *in the target repo*, why are you adding a version number here?

Please add a comment above the line of code explaining all of this.

.gitignore
inteon: done

modules/tools/00_mod.mk
inteon: I added a comment

wallrj (approved):
Thanks.

/lgtm
/approve

PR #7, opened 2023-12-18
Fix bug in self-upgrade script

inteon:
There is a bug in the self-upgrade script, the script succeeds but does not create a PR (see https://github.com/inteon/approver-policy/actions/runs/7251155494/job/19752903084).
This is because the GH API requires the head parameter to also contain the organisation where the branch lives.

This is an example of a successful run and the created PR:
This is an example run: https://github.com/inteon/approver-policy/actions/runs/7251155494
The created PR: https://github.com/inteon/approver-policy/pull/1

### DIFF
--- a/modules/repository-base/base/.github/workflows/make-self-upgrade.yaml
+++ b/modules/repository-base/base/.github/workflows/make-self-upgrade.yaml
@@ -67,7 +67,7 @@ jobs:
             const pulls = await github.rest.pulls.list({
               owner: owner,
               repo: repo,
-              head: 'self-upgrade',
+              head: owner + ':self-upgrade',
               base: 'main',
               state: 'open',
             });


wallrj (changes requested):
Please fill in the PR description.
What bug?
How do you know this fixes it?

inteon:
> Please fill in the PR description. What bug? How do you know this fixes it?

I'm testing it here: https://github.com/cert-manager/approver-policy/pull/322

inteon:
@wallrj interesting finding:
When GH detects that you try to create/ alter a GH action from a GH action, it does not allow you: 
```
! [remote rejected] self-upgrade -> self-upgrade (refusing to allow a GitHub App to create or update workflow `.github/workflows/make-self-upgrade.yaml` without `workflows` permission)
```

Which is maybe not a bad thing. but it does mean that not everything will be auto-upgradable. As an alternative, we can add a GH token as a secret and use a bot account which creates forks to upgrade changes (sounds more complex, but maybe something prow could do).
For now, I guess we have to accept that changes to the GH actions cannot be auto-upgraded.

inteon:
Appart from the issue explained above, this fix works.
This is an example run: https://github.com/inteon/approver-policy/actions/runs/7251155494
The created PR: https://github.com/inteon/approver-policy/pull/1

wallrj (approved):
Thanks @inteon 

/lgtm
/approve

PR #9, opened 2023-12-18
Signoff self-upgrade commit messages

inteon:
the `--signoff` flag was missing, causing the dco check to fail.


wallrj (approved):
/lgtm
/approve

PR #13, opened 2024-01-02
Upgrade the distroless base image to the debian12 variant

inteon:
This upgrades all libraries that use the makefile-modules (currently just approver-policy) to use the debian12 distroless base image.


SgtCoDFish (approved):
/lgtm
/approve

PR #14, opened 2024-01-02
Set default value for the artifacts directory

inteon:
Adds the ARTIFACTS variable with a default value set.
This ARTIFACTS variable is set by prow and the contents of this directory are uploaded after the test (see https://docs.prow.k8s.io/docs/components/pod-utilities/).

### DIFF
--- a/modules/repository-base/base/Makefile
+++ b/modules/repository-base/base/Makefile
@@ -67,7 +67,15 @@ GITEPOCH := $(shell git show -s --format=%ct HEAD)
 
 bin_dir := _bin
 
-$(bin_dir) $(bin_dir)/scratch:
+# The ARTIFACTS environment variable is set by the CI system to a directory
+# where artifacts should be placed. These artifacts are then uploaded to a
+# storage bucket by the CI system (https://docs.prow.k8s.io/docs/components/pod-utilities/).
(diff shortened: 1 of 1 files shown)


modules/repository-base/base/Makefile
```diff
@@ -65,6 +65,8 @@ GITEPOCH := $(shell git show -s --format=%ct HEAD)
 # Global variables and dirs      #
 ##################################
 
+ARTIFACTS ?= _artifacts
```
wallrj, line 68: Add a comment explaining what this variable is used for.
Include the link to that Prow documentation.

`ARTIFACTS` is not used anywhere in *this* repo, so where is it used / how should it be used?
* https://github.com/search?q=repo%3Acert-manager%2Fmakefile-modules%20ARTIFACTS&type=code

Also consider changing the default to `_bin/artifacts` so that all the generated files are in one place and so that I don't have to add another line to the `.gitignore` file of each project.

wallrj (changes requested):

modules/repository-base/base/Makefile
inteon: I added a comment explaining what this variable is used for.

SgtCoDFish (approved):
/lgtm
/approve

PR #17, opened 2024-01-08
Include version in the cert-manager yaml file name

inteon:
To prevent a stale yaml file from being used, we should include the `cert_manager_version` variable in the filename.
This will cause the file to be re-downloaded when `cert_manager_version` changes.


SgtCoDFish (approved):
/lgtm
/approve

👍

PR #18, opened 2024-01-08
Add images-learn-sha target and improve sha tools

inteon:
- improves image hash checking (based on https://github.com/cert-manager/cert-manager/pull/6440)
- removes `tools-learn-sha` from help of projects that import the `tools` module; because the target should only be used in this repo.
- adds `images-learn-sha` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,15 +29,32 @@ SHELL := /usr/bin/env bash
 .SUFFIXES:
 FORCE:
 
+HOST_OS ?= $(shell uname -s | tr A-Z a-z)
+HOST_ARCH ?= $(shell uname -m)
+ifeq (x86_64, $(HOST_ARCH))
+	HOST_ARCH = amd64
+endif
+
 bin_dir := _bin
 
-$(bin_dir):
+$(bin_dir) $(bin_dir)/scratch:
 	mkdir -p $@
 
-include modules/tools/00_mod.mk
+include modules/**/00_mod.mk
+
+.PHONY: images-learn-sha
+images-learn-sha: $(bin_dir) | $(NEEDS_CRANE)
+	rm -rf ./$(bin_dir)/downloaded/images/
+	mkdir -p ./$(bin_dir)/scratch/
+
+	IMAGES_AMD64="$(images_amd64)" \
+	IMAGES_ARM64="$(images_arm64)" \
+	CRANE=$(CRANE) \
+		./scripts/images_learn_sha.sh
 
 .PHONY: help
 help: ## Show this help
 	@echo "Usage: make [target] ..."
 	@echo
 	@echo "make tools-learn-sha"
+	@echo "make images-learn-sha"
--- a/modules/kind/00_mod.mk
+++ b/modules/kind/00_mod.mk
@@ -15,5 +15,7 @@
 images_amd64 ?=
 images_arm64 ?=
 
-images_amd64 += docker.io/kindest/node:v1.27.3@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -29,15 +29,48 @@ SHELL := /usr/bin/env bash
...
+	for image in $(images_arm64); do \
+		image_no_digest=$$(echo -n "$${image}" | cut -d@ -f1); \
+		find=$$(echo -n "$${image}" | cut -d@ -f2); \
+		replace=$$($(CRANE) digest --platform "linux/arm64" "$${image_no_digest}"); \
+		echo "s|$${find}|$${replace}|g" >> "$(LEARN_FILE)"; \
+	done
+
+	while read p; do \
+		find ./modules/ -maxdepth 2 -name "00_mod.mk" -type f -exec sed -i "$$p" {} \;; \
+	done <"$(LEARN_FILE)"
```
SgtCoDFish, line 69: suggestion: To me this is complicated enough that it would benefit from being a standalone bash script which `make images-learn-sha` simply calls.

It'd be easier to read (fewer `\`s and `;`s) and debug (e.g. adding shellcheck later).

Plus this make target is phony but modifies named files. That definitely feels like more of a bash script behaviour to me. What do you think?

modules/kind/00_mod.mk
```diff
@@ -15,5 +15,7 @@
 images_amd64 ?=
 images_arm64 ?=
 
-images_amd64 += docker.io/kindest/node:v1.27.3@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde
-images_arm64 += docker.io/kindest/node:v1.27.3@sha256:de0b3dfe848ccf07e24f4278eaf93edb857b6231b39773f46b36a2b1a6543ae9
+kind_version := v1.27.3
+
+images_amd64 += docker.io/kindest/node:$(kind_version)@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde
+images_arm64 += docker.io/kindest/node:$(kind_version)@sha256:de0b3dfe848ccf07e24f4278eaf93edb857b6231b39773f46b36a2b1a6543ae9
```
SgtCoDFish, line 21: question: From previous discussions, I remember Kind being a bit weird with container images. The [latest release](https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0) says explicitly that the exact images listed in the release must be used.

We've had issues with this in cert-manager too.

Does this script automatically update kind images using crane? If so, is it possible / desirable for us to not do that?

SgtCoDFish (commented):
Couple of bits on this one - I like the idea!

modules/kind/00_mod.mk
inteon: Thanks for the pointer, I did not realise kind has its own special way of tagging images.
I'll try to not update the kind images in the script.

Makefile
inteon: I moved the logic to an bash script.

modules/kind/00_mod.mk
inteon: I added an exception for `docker.io/kindest/node` to the script.

scripts/images_learn_sha.sh
```diff
@@ -0,0 +1,57 @@
...
+    image_no_digest=$(echo -n "$image" | cut -d@ -f1)
+    find=$(echo -n "$image" | cut -d@ -f2)
+    replace=$($CRANE digest --platform "linux/arm64" "$image_no_digest")
+
+    if [ "$find" == "$replace" ]; then
+        continue
+    fi
+
+    learn_data+=("s|$find|$replace|g\n")
+done
```
SgtCoDFish, line 46: comment: this could be a function, but that's complicated enough in bash that it's fine to copy here I think!

SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #20, opened 2024-01-11
Upgrade kind k8s to v1.29.0

inteon:


modules/kind/00_mod.mk
```diff
@@ -15,7 +15,14 @@
 images_amd64 ?=
 images_arm64 ?=
 
-kind_version := v1.27.3
+kind_version := v1.29.0
 
-images_amd64 += docker.io/kindest/node:$(kind_version)@sha256:9dd3392d79af1b084671b05bcf65b21de476256ad1dcc853d9f3b10b4ac52dde
-images_arm64 += docker.io/kindest/node:$(kind_version)@sha256:de0b3dfe848ccf07e24f4278eaf93edb857b6231b39773f46b36a2b1a6543ae9
+# Goto https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0 and find the
```
SgtCoDFish, line 20: suggestion: This URL is correct now but won't be in the future. Future maintainers might copy+paste without noticing there's a version number in there.

```suggestion
# Goto https://github.com/kubernetes-sigs/kind/releases/tag/<KIND-VERSION> and find the
```

SgtCoDFish (changes requested):

modules/kind/00_mod.mk
```diff
@@ -15,7 +15,14 @@
 images_amd64 ?=
 images_arm64 ?=
 
-kind_version := v1.27.3
+kind_version := v1.29.0
```
SgtCoDFish, line 18: suggestion (non-blocking): I think this variable is confusingly named because it makes me think we're talking about the version of the `kind` tool.

This is the version of k8s we'll be using with `kind`, right?

Might be worth changing to `kind_k8s_version` or something, but that'd be a separate PR!

SgtCoDFish (approved):
/lgtm
/approve

Looks good, cheers!

modules/kind/00_mod.mk
inteon: Yes, the name is confusing, I'll create a PR.

PR #24, opened 2024-01-12
Add an autobumper for the vendored Go version

SgtCoDFish:
This only attempts to bump patch versions, leaving us free to make the choice of when to bump the minor version


inteon:
Thanks, looks very promising!
/approve
/lgtm

PR #25, opened 2024-01-12
Fix indentation in GH action

inteon:
I missed this in my review of #24.


inteon:
self-lgtm'ng because this is a very simple fix.

PR #26, opened 2024-01-15
Fix autobumper script to use gnused

SgtCoDFish:
The CLI tool is different on macOS! Also fixes the cleanup command and tweaks the output


inteon (commented):
/approve
/lgtm

PR #28, opened 2024-01-15
Add goreleaser and syft as tools

inteon:
Adding these tools so we can use them for releases.

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -90,6 +90,10 @@ TOOLS += oras=v1.1.0
 TOOLS += ginkgo=$(shell [[ -f go.mod ]] && awk '/ginkgo\/v2/ {print $$2}' go.mod || echo "v2.13.2")
 # https://pkg.go.dev/github.com/cert-manager/klone?tab=versions
 TOOLS += klone=v0.0.3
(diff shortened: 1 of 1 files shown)


modules/tools/00_mod.mk
```diff
@@ -262,6 +266,8 @@ GO_DEPENDENCIES += boilersuite=github.com/cert-manager/boilersuite
 GO_DEPENDENCIES += gomarkdoc=github.com/princjef/gomarkdoc/cmd/gomarkdoc
 GO_DEPENDENCIES += oras=oras.land/oras/cmd/oras
 GO_DEPENDENCIES += klone=github.com/cert-manager/klone
+GO_DEPENDENCIES += goreleaser=github.com/goreleaser/goreleaser
+GO_DEPENDENCIES += syft=github.com/anchore/syft/cmd/syft
```
SgtCoDFish, line 270: note: syft uses goreleaser for its own releases, and it defines variables in the linker which we'll leave unset since we're using go install. I think that's fine: https://github.com/anchore/syft/blob/0748413d6bd5f668a9c66ab5632d320df5e28458/.goreleaser.yaml#L27

SgtCoDFish (approved):
/lgtm
/approve

PR #30, opened 2024-01-15
Helm "artifacthub.io/prerelease" annotation

inteon:
This PR adds a step to the Helm chart creation process that sets the "artifacthub.io/prerelease" annotation in the Chart.yaml file.

### DIFF
--- a/modules/repository-base/base/Makefile
+++ b/modules/repository-base/base/Makefile
@@ -58,6 +58,7 @@ endif
 ##################################
 
 VERSION ?= $(shell git describe --tags --always --match='v*' --abbrev=14 --dirty)
+IS_PRERELEASE := $(shell git describe --tags --always --match='v*' --abbrev=0 | grep -q '-' && echo true || echo false)
 GITCOMMIT := $(shell git rev-parse HEAD)
 GITEPOCH := $(shell git show -s --format=%ct HEAD)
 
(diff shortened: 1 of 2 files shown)


inteon:
I tried this on the approver-policy repo, and it seems to work locally.
/unhold

modules/repository-base/base/Makefile
```diff
@@ -58,6 +58,7 @@ endif
 ##################################
 
 VERSION ?= $(shell git describe --tags --always --match='v*' --abbrev=14 --dirty)
+IS_PRERELEASE := $(shell echo "$(VERSION)" | grep -q '-' && echo true || echo false)
```
SgtCoDFish, line 61: suggestion: Is this safe? Imagining we're releasing a new version `v50.0.0`; if part of one of the modules started to (accidentally) make the git state dirty during a release, this would make the version show as a prerelease.

E.g.

```console
$ git tag -s -m"v50" v50.0.0
$ echo test > hello && git add hello
$ git describe --tags --always --match='v*' --abbrev=14 --dirty
v50.0.0-dirty
```

Because of the dash, this would be marked as a prerelease version. It's easy to imagine how this could happen.

There are a lot of ways to solve this - we probably should cancel a release if the git tag is dirty, we could change this parsing to check for `-alpha` or `-beta`, we could not use `--dirty` in VERSION or we could have `RELEASE_VERSION` which doesn't use `--dirty`.

What do you think? Is this a concern?

SgtCoDFish (approved):
/lgtm
/approve

Makes sense to me, good to have this in as many places as possible!

PR #31, opened 2024-01-15
Set empty 'shared_generate_targets' and 'shared_verify_targets_dirty' variables in the generate-verify module

inteon:
Simplifies the module and prevents warnings in case these variables have not been set.


SgtCoDFish (approved):
/lgtm
/approve

PR #32, opened 2024-01-15
Split oci-image module into 'executable' and 'oci-image' modules

inteon:
See https://github.com/cert-manager/cmctl for a working example.

### DIFF
--- a/modules/executable/01_mod.mk
+++ b/modules/executable/01_mod.mk
@@ -1,0 +1,110 @@
+# Copyright 2023 The cert-manager Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+exe_platforms := all
+
+ifndef bin_dir
+$(error bin_dir is not set)
+endif
+
+ifndef build_names
+ifndef exe_build_names
+$(error build_names and exe_build_names are not set)
+endif
+endif
+
+all_exe_build_names := $(sort $(build_names) $(exe_build_names))
+
+fatal_if_undefined = $(if $(findstring undefined,$(origin $1)),$(error $1 is not set))
+
+define check_variables
+$(call fatal_if_undefined,go_$1_ldflags)
+$(call fatal_if_undefined,go_$1_source_path)
+endef
+
+$(foreach build_name,$(all_exe_build_names),$(eval $(call check_variables,$(build_name))))
+
+ifdef exe_build_names
+$(call fatal_if_undefined,gorelease_file)
+endif
+
+##########################################
+
+RELEASE_DRYRUN ?= false
+
+CGO_ENABLED ?= 0
+
+run_targets := $(all_exe_build_names:%=run-%)
+build_targets := $(all_exe_build_names:%=$(bin_dir)/bin/%)
+
+$(bin_dir)/bin:
+	mkdir -p $@
+
+.PHONY: $(run_targets)
+ARGS ?= # default empty
+## Directly run the go source locally.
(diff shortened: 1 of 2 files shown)


modules/executable/01_mod.mk
```diff
@@ -0,0 +1,110 @@
...
+
+.PHONY: $(run_targets)
+ARGS ?= # default empty
+## Directly run the go source locally.
+## @category [shared] Build
+$(run_targets): run-%: | $(NEEDS_GO)
+	CGO_ENABLED=$(CGO_ENABLED) \
+	$(GO) run \
+		-ldflags '$(go_$*_ldflags)' \
+		$(go_$*_source_path) $(ARGS)
```
SgtCoDFish, line 62: suggestion: I was actually gonna propose removing this from the oci-image module. I don't really think there's a huge amount of value in it - better IMO to build + run the binary so you're testing something closer to what you'd release. Also prevents us having to keep things up to date in the future (i.e. making sure the go build and go run have the same GOEXPERIMENT envvars, etc)

What do you think?

modules/executable/01_mod.mk
```diff
@@ -0,0 +1,110 @@
...
+ifeq ($(RELEASE_DRYRUN),true)
+	$(eval extra_args := $(extra_args) --skip=announce,publish,validate,sign)
+endif
+
+	SOURCE_DATE_EPOCH=$(GITEPOCH) \
+	CGO_ENABLED=$(CGO_ENABLED) \
+	$(GORELEASER) release \
+		$(extra_args) \
+		--fail-fast \
+		--config=$(CURDIR)/$(go_releaser_path).goreleaser_config.yaml
```
SgtCoDFish, line 110: suggestion: If we're using goreleaser to release, do you think we should be using it instead of `go build` for local builds too - so the build configuration is defined in a single place?

modules/oci-image/01_mod.mk
```diff
@@ -54,41 +54,19 @@ $(foreach build_name,$(build_names),$(eval $(call check_variables,$(build_name))
...
-
-.PHONY: $(run_targets)
-ARGS ?= # default empty
-## Run a controller from your host.
-## @category [shared] Build
-$(run_targets): run-%: | $(NEEDS_GO)
-	CGO_ENABLED=$(CGO_ENABLED) \
-	$(GO) run \
-		-ldflags '$(go_$*_ldflags)' \
-		$(go_$*_source_path) $(ARGS)
```
SgtCoDFish, line 88: comment: I really like removing this and making it a separate module!

modules/oci-image/01_mod.mk
```diff
@@ -54,41 +54,19 @@ $(foreach build_name,$(build_names),$(eval $(call check_variables,$(build_name))
...
-
 .PHONY: $(oci_build_targets)
-## Build the oci image.
+## Build the OCI image.
```
SgtCoDFish, line 69: comment: It'd be great if this targets build args (CGO_ENABLED, ldflags, etc) could be more closely synchronised with those used by goreleaser / `go build` - again, aiming for a single definition so we don't test and release different things.

Not sure if it's worth the effort to do that now, though.

SgtCoDFish (commented):
Small review as requested!

modules/oci-image/01_mod.mk
inteon: That is what I'm trying to do here: https://github.com/cert-manager/makefile-modules/pull/32/files/51f9b6cf6383621401c0cacc8818d28525657b80#diff-1a68d7e0cc16504904116bcd1028f40525f8d8ea25feb5ee5559d47fa765f1e8R74-R84

I basically add the CGO_ENABLED, ldflags etc. to the gorelease config yaml file before calling gorelease release.

SgtCoDFish (approved):
/lgtm
/approve

I don't think any of the things I raised are blockers for this PR and I need to raise several followup PRs to these modules which require this to be unblocked

modules/oci-image/01_mod.mk
SgtCoDFish: I think it'd be safer to also build locally using goreleaser (i.e. don't use `go build -o $@` locally, use `$(GORELEASER) build --snapshot --clean --single-target --output $@`)

If we `go build` and use that locally (or in tests, whatever) then don't use that same built binary for a release, there are a bunch of problems that could be be introduced - say if there's a problem with setting an ldflag in the goreleaser config.

PR #33, opened 2024-01-15
BUGFIX: self-upgrade should also commit new files

inteon:
I noticed that in https://github.com/cert-manager/approver-policy/pull/345 a newly created file is missing from the PR.
This is due to a bug in the self-upgrade script.
This PR fixes that bug.


SgtCoDFish (approved):
/lgtm
/approve

Seems to be a clear bugfix!

ThatsMrTalbot (approved):

PR #37, opened 2024-01-17
Update SHELLFLAGS and add comment that explains where these values come from

inteon:
Thanks to https://tech.davis-hansson.com/p/make/ for the article.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,11 @@
 
 ###################################################################################
 
+# See https://tech.davis-hansson.com/p/make/ for more information
+# about these initial lines.
 MAKEFLAGS += --warn-undefined-variables --no-builtin-rules
 SHELL := /usr/bin/env bash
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,9 +21,11 @@
 
 ###################################################################################
 
+# See https://tech.davis-hansson.com/p/make/ for more information
+# about these initial lines.
 MAKEFLAGS += --warn-undefined-variables --no-builtin-rules
 SHELL := /usr/bin/env bash
-.SHELLFLAGS := -uo pipefail -c
+.SHELLFLAGS := -eu -o pipefail -c
```
SgtCoDFish, line 28: I wish I could remember why we don't have errexit set in the cert-manager makefile. I feel like it broke something, but don't have time to dig into it 🙃 

If it works here though, all the better!

SgtCoDFish (approved):
/lgtm
/approve

PR #38, opened 2024-01-17
Remove extra_*_targets variables

inteon:
In https://github.com/cert-manager/makefile-modules/pull/31, I tried to simplify these variables.
However, at this point I do not know why we have separate `extra_generate_targets` and `shared_generate_targets` variables.


SgtCoDFish (approved):
/lgtm
/approve

Haven't tested but can review followups

PR #39, opened 2024-01-18
feat: use helm-tool to generate docs

ThatsMrTalbot:
Allows using helm-tool to generate docs if desired, change is gated behind a variable to prevent current repos from using it unintentionally

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -68,11 +68,24 @@ $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_d
 ## @category [shared] Helm Chart
 helm-chart: $(helm_chart_archive)
 
+ifdef helm_docs_use_helm_tool
+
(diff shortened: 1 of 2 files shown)


modules/helm/helm.mk
```diff
@@ -68,11 +68,24 @@ $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_d
 ## @category [shared] Helm Chart
 helm-chart: $(helm_chart_archive)
 
+ifdef helm_docs_use_helm_tool
+
+helm_tool_header_search ?= ^\#\#\\s+Parameters *$$
```
inteon, line 73: Could we use `<!-- AUTO-GENERATED -->` instead?

inteon: Now, there would be no indication that the table is auto-generated.

ThatsMrTalbot: Yea, sounds good to me. Updated

inteon (commented):
/approve
/lgtm

PR #42, opened 2024-01-18
Add linux/arm64 support

SgtCoDFish:
This is common when using containers on macOS for development, but linux/arm64 is common enough anyway.


SgtCoDFish:
Adding self approval given the LGTM!

PR #43, opened 2024-01-18
Add ability to configure GOEXPERIMENT for builds

SgtCoDFish:
It'd be great to be able to set things like `loopvar` in the future using this!


SgtCoDFish:
(adding self approval given the LGTM)

PR #46, opened 2024-01-19
Tweak checkhash script

SgtCoDFish:
I noticed that this script wouldn't exit gracefully when I ran into an edge case using makefiles on a previously unsupported platform (linux/arm64).

This should make the script more resilient if it's called with a missing arg (and it IMO simplifies the control flow a little)


inteon (commented):
Looks good!
/approve
/lgtm

PR #50, opened 2024-01-22
Make oci_platforms conditional

SgtCoDFish:
This is useful if, say, building a test locally using only the local arch.


SgtCoDFish:
(adding self approval since I've got an LGTM)

PR #51, opened 2024-01-23
feat: allow additional tools to be defined

ThatsMrTalbot:
Allows ADDITIONAL_TOOLS and ADDITIONAL_GO_DEPENDENCIES to be used to include more tools reusing the same downloading logic

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -109,6 +109,10 @@ KUBEBUILDER_ASSETS_VERSION=1.28.3
 TOOLS += etcd=$(KUBEBUILDER_ASSETS_VERSION)
 TOOLS += kube-apiserver=$(KUBEBUILDER_ASSETS_VERSION)
 
+# Additional tools can be defined to reuse the tooling in this file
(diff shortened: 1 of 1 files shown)


modules/tools/00_mod.mk
```diff
@@ -109,6 +109,10 @@ KUBEBUILDER_ASSETS_VERSION=1.28.3
 TOOLS += etcd=$(KUBEBUILDER_ASSETS_VERSION)
 TOOLS += kube-apiserver=$(KUBEBUILDER_ASSETS_VERSION)
 
+# Additional tools can be defined to reuse the tooling in this file
+ADDITIONAL_TOOLS ?=
+TOOLS += $(ADDITIONAL_TOOLS)
```
inteon, line 114: Another option is to use `TOOLS ?= ` here and in the other place that adds a tool.

ThatsMrTalbot: That was another thought I had, I did worry it would pick up stray environment vars though. I dont mind either way tbh

PR #54, opened 2024-01-24
Explicitly define goreleaser targets instead of arch,os matrix

inteon:
As requested by @SgtCoDFish, instead of relying on goreleaser to ignore nonsensical os,arch combinations, this PR instead explicitly defines all combinations that we want to build.

### DIFF
--- a/modules/executable/01_mod.mk
+++ b/modules/executable/01_mod.mk
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-exe_operating_systems ?= darwin,linux,windows
-exe_architectures ?= amd64,arm,arm64,ppc64le,s390x
+exe_targets ?= darwin_amd64_v1,darwin_arm64,linux_amd64_v1,linux_arm_6,linux_arm64,linux_ppc64le,linux_s390x,windows_amd64_v1,windows_arm_6,windows_arm64
 
 ifndef bin_dir
 $(error bin_dir is not set)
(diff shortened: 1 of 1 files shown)


modules/executable/01_mod.mk
```diff
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-exe_operating_systems ?= darwin,linux,windows
-exe_architectures ?= amd64,arm,arm64,ppc64le,s390x
+exe_targets ?= darwin_amd64_v1,darwin_arm64,linux_amd64_v1,linux_arm_6,linux_arm64,linux_ppc64le,linux_s390x,windows_amd64_v1,windows_arm_6,windows_arm64
```
SgtCoDFish, line 15: suggestion: we previously built for armv7, not armv6 as we have here: https://github.com/cert-manager/cert-manager/blob/876e386ee905aa86e2466c287e654613b0426927/make/cmctl.mk#L41

I'd vote to keep things the same as we had before in terms of which version we target

SgtCoDFish, line 15: suggestion: this PR also adds 32-bit ARM for Windows, which we didn't previously build. I don't think we should compile this - we've discussed before about how our use of unsized `int` in API types might mean inconsistencies between 32-bit and 64-bit platforms.

I think we should continue to build linux/armv7 because we already did - but I don't think we should add new 32-bit platforms unless we explicitly get asked.

I'm fine with adding windows/arm64 though, since that's only going to get more common and doesn't have the same pitfalls

SgtCoDFish (changes requested):
Couple of suggestions!

modules/executable/01_mod.mk
inteon: updated

inteon: updated

SgtCoDFish (approved):
/lgtm
/approve

Thanks 🚀

PR #56, opened 2024-01-24
Fix loading kind images on macOS

SgtCoDFish:
This image is run outside containers, so `runtime.GOOS` is `darwin` on macOS and no image will ever be found.

The kind target environment on M1 is `linux/arm64`, and the target environment on Intel macs would be `linux/amd64` - no need for `GOOS`.


inteon:
Makes sense, there are no darwin containers yet.
/approve
/lgtm

PR #57, opened 2024-01-26
fix: defer dependency expansion of generate and verify targets

ThatsMrTalbot:
If a module adds a variable to shared_generate_targets after the generate module has been parsed, it is not include in `make generate` or `make verify`. This change uses a second expansion to defer the evaluation of those dependencies to the end

### DIFF
--- a/modules/generate-verify/02_mod.mk
+++ b/modules/generate-verify/02_mod.mk
@@ -15,7 +15,7 @@
 .PHONY: generate
 ## Generate all generate targets.
 ## @category [shared] Generate/ Verify
-generate: $(shared_generate_targets)
+generate: $$(shared_generate_targets)
 
 verify_script := $(dir $(lastword $(MAKEFILE_LIST)))/util/verify.sh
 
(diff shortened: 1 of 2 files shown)


modules/generate-verify/02_mod.mk
```diff
@@ -15,17 +15,19 @@
...
-generate: $(shared_generate_targets)
+generate: $$(shared_generate_targets)
 
 verify_script := $(dir $(lastword $(MAKEFILE_LIST)))/util/verify.sh
 
 # Run the supplied make target argument in a temporary workspace and diff the results.
 verify-%: FORCE
 	$(verify_script) $(MAKE) -s $*
 
+verify_generated_targets = $(shared_generate_targets:%=verify-%)
```
inteon, line 26: @ThatsMrTalbot what is the reason this is moved to a new variable?
Does that not cause the variable to be evaluated directly (not on second pass)?

ThatsMrTalbot: The `$$` syntax does not like the "%", it thinks its part of a target wildcard. Since `=` variables are not evaluated until they are expanded, this solves the issue

PR #68, opened 2024-02-07
feat: allow custom base image

ThatsMrTalbot:
Allow custom base images to be specified using 
```
oci_<name>_base_image_flavor := custom
oci_<name>_base_image := my-custom-base-image
````

### DIFF
--- a/modules/oci-image/01_mod.mk
+++ b/modules/oci-image/01_mod.mk
@@ -45,6 +45,8 @@ ifeq ($(oci_$1_base_image_flavor),static)
     oci_$1_base_image := $(base_image_static)
 else ifeq ($(oci_$1_base_image_flavor),csi-static)
     oci_$1_base_image := $(base_image_csi-static)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/oci-image/01_mod.mk [hunk 1]
```diff
@@ -45,6 +45,8 @@ ifeq ($(oci_$1_base_image_flavor),static)
     oci_$1_base_image := $(base_image_static)
 else ifeq ($(oci_$1_base_image_flavor),csi-static)
     oci_$1_base_image := $(base_image_csi-static)
+else ifeq ($(oci_$1_base_image_flavor),custom)
+	$(call fatal_if_undefined,oci_$1_base_image)
```
inteon, line 49: @ThatsMrTalbot
I think this should have been `$$(call fatal_if_undefined,oci_$1_base_image)`
Right now, this breaks eg. approver-policy with the following error:
```console
make/_shared/oci-image/01_mod.mk:75: *** oci_manager_base_image is not set.  Stop.
```

PR #73, opened 2024-02-12
Add 'crds.enabled' and 'crds.keep' options to generated CRDs

inteon:
Implements https://github.com/cert-manager/cert-manager/pull/6719 for all the Makefile module projects.

See https://github.com/cert-manager/approver-policy/pull/376 for an example.


modules/helm/crd.template.yaml
```diff
@@ -0,0 +1,12 @@
+# START crd {{- if or .Values.crds.enabled }}
```
ThatsMrTalbot, line 1: Since the comments are in YAML they will be visible if a user does a helm template
The comments are also outside the `if` statement, meaning they always print regardless of if the CRD does.

Is this desired behaviour?

modules/helm/crd.template.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: "SHOULD-REPLACE"
+  # START Helm policy annotations {{- if .Values.crds.keep }}
+  annotations:
+    helm.sh/resource-policy: keep
+  # END Helm policy annotations {{- end }}
+  labels: {}
```
ThatsMrTalbot, line 10: We don't have the recommended helm labels here:
https://helm.sh/docs/chart_best_practices/labels/#standard-labels

modules/helm/crd.template.yaml
inteon: Yes, I'm creating a template that is also valid YAML so I can still use tools like yq.

modules/helm/crd.template.yaml
inteon: I set a comment in the linked example to template these labels:
https://github.com/cert-manager/approver-policy/pull/376/files#diff-4c5b6a82b6742542e1cc2458d47956d7fb1858e433d78b044a4cde6149a422b6R10
and
https://github.com/cert-manager/approver-policy/pull/376/files#diff-6610634eaa887c0abfe1832cf1ecd9cf3d6b4be7d1766cdd279e4e58d27d685aR50-R53

inteon:
@ThatsMrTalbot I updated this PR and the approver-policy PR: made the templates more readable (stopped using those comments).

PR #81, opened 2024-03-01
feat: add ability to push and sign images separately

ThatsMrTalbot:
This creates targets to push without signing as well as targets to just sign the image.

It also adds adds the ability to push to multiple registries by specifying multiple values for `oci_<name>_image_name`

### DIFF
--- a/modules/oci-image/01_mod.mk
+++ b/modules/oci-image/01_mod.mk
@@ -80,8 +80,13 @@ RELEASE_DRYRUN ?= false
 CGO_ENABLED ?= 0
 GOEXPERIMENT ?=  # empty by default
 
+COSIGN_FLAGS ?= # empty by default
+OCI_SIGN_ON_PUSH ?= true
+
 oci_build_targets := $(build_names:%=oci-build-%)
 oci_push_targets := $(build_names:%=oci-push-%)
+oci_push_no_sign_targets := $(build_names:%=oci-push-no-sign-%)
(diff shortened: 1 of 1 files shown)


modules/oci-image/01_mod.mk
```diff
@@ -138,6 +143,25 @@ $(oci_build_targets): oci-build-%: | $(NEEDS_KO) $(NEEDS_GO) $(NEEDS_YQ) $(bin_d
...
+endef
+
+
+
+
+
+# Functions for pushing and signing. We have a few targets that push/sign, this
+# use of functions means we can define the commands once.
+oci_digest = $(shell head -1 $(CURDIR)/$(oci_layout_path_$1).digests)
+oci_push_command = $(foreach oci_image_name,$(oci_$1_image_name),$(CRANE) push "$(oci_layout_path_$1)" "$(oci_image_name):$(oci_$1_image_tag);")
```
inteon, line 162: I just realised that we do expect oci_image_name to be a single image in some cases (eg. the Helm chart).
I do think however that that is only in the non-makefile modules part of the Helm chart logic.

inteon: So we can probably ignore it here.

ThatsMrTalbot: Could make it just pick the first image? Seems like a sane default

inteon (approved):
LGTM
/lgtm

PR #88, opened 2024-03-12
Run 'go mod tidy'

inteon:
Was missing from https://github.com/cert-manager/makefile-modules/pull/87


inteon:
Self-approving this fix.

PR #91, opened 2024-03-14
feat: split oci-image module to oci-build and oci-publish

ThatsMrTalbot:
This splits out the `oci-image` module into `oci-build` and `oci-publish`. The general idea being that projects can use one or the other if required, using a different tool to build images if needed, or using different mechanism to publish the built images.

Testing:
This was tested by starting a local registry with `crane registry serve` and crafting a minimal config to for the makefile to function. 

Other improvements:
- The image_tool is built once and stored in `_bin/tools` along with other tools
- The `oci-load-%` command is omitted if the kind module is not included
- Fixed `docker-tarball-%` is now building the correct arch on m3 macbooks (previously it would build amd64 for som reason)


modules/oci-publish/02_mod.mk
```diff
@@ -0,0 +1,124 @@
...
+
+$(foreach build_name,$(build_names),$(eval $(call default_per_build_variables,$(build_name))))
+
+# Validate variables per build_names entry
+#
+# $1 - build_name
+define check_per_build_variables
+$(call fatal_if_undefined,oci_digest_path_$1)
+$(call fatal_if_undefined,oci_layout_path_$1)
+$(call fatal_if_undefined,oci_image_name_$1)
```
inteon, line 53: Can we keep the old name here, that will make it easier to migrate and test these changes? `oci_$1_image_name`

modules/oci-build/01_mod.mk
```diff
@@ -0,0 +1,198 @@
...
+# Derive config based on user config
+# 
+# - oci_layout_path_$(build_name) = path that the OCI image will be saved in OCI layout directory format
+# - oci_digest_path_$(build_name) = path to the file that will contain the digests
+# - ko_config_path_$(build_name) = path to the ko config file
+# - docker_tarball_path_$(build_name) = path that the docker tarball that the docker-tarball-$(build_name) will produce
+$(foreach build_name,$(build_names),$(eval oci_layout_path_$(build_name) := $(bin_dir)/scratch/image/oci-layout-$(build_name).$(oci_$(build_name)_image_tag)))
+$(foreach build_name,$(build_names),$(eval oci_digest_path_$(build_name) := $(CURDIR)/$(oci_layout_path_$(build_name)).digests))
+$(foreach build_name,$(build_names),$(eval ko_config_path_$(build_name) := $(CURDIR)/$(oci_layout_path_$(build_name)).ko_config.yaml))
+$(foreach build_name,$(build_names),$(eval docker_tarball_path_$(build_name) := $(CURDIR)/$(oci_layout_path_$(build_name)).docker.tar))
```
inteon, line 128: I think we can move these definitions to a 00_mod.mk file and move the oci-publish/02_mod.mk file to oci-publish/01_mod.mk

ThatsMrTalbot: This requires tools like KO, which is done in a 00_mod.mk, since `o` becomes before `t` this would get included before tools so the `NEEDS_KO` like vars would not be defined

ThatsMrTalbot: Do we also want to move the validation? It makes sense to only define these after validation

inteon: Yes, we can move everything. All these variables are expected to be defined in the make/00_mod.mk file anyways.

ThatsMrTalbot: Moved the vars and renamed the oci-publish/02_mod.mk file to oci-publish/01_mod.mk

modules/oci-build/01_mod.mk
```diff
@@ -0,0 +1,198 @@
...
+		$(YQ) '.builds[0].main = "$(go_$1_main_dir)"' | \
+		$(YQ) '.builds[0].env[0] = "CGO_ENABLED=$(cgo_enabled_$1)"' | \
+		$(YQ) '.builds[0].env[1] = "GOEXPERIMENT=$(goexperiment_$1)"' | \
+		$(YQ) '.builds[0].ldflags[0] = "-s"' | \
+		$(YQ) '.builds[0].ldflags[1] = "-w"' | \
+		$(YQ) '.builds[0].ldflags[2] = "{{.Env.LDFLAGS}}"' \
+		> $(CURDIR)/$(oci_layout_path_$1).ko_config.yaml
+
+ko-config-$1: $(ko_config_path_$1:$(CURDIR)/%=%)
+endef
```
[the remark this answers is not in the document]
ThatsMrTalbot: Fair enough, I have made the target `.PHONY` so it always runs

modules/oci-publish/02_mod.mk
ThatsMrTalbot: I don't know why I renamed it, its now the old name

inteon (commented):
LGTM, building seems to work for approver-policy
/approve
/lgtm

PR #92, opened 2024-03-15
fix: make shasum work on darwin without coreutils installed

ThatsMrTalbot:


inteon (approved):
LGTM.
I did not test because I have no mac, but I trust that it works.
/approve
/lgtm

PR #93, opened 2024-03-17
remove extra ;

abvaidya:
oci_push_command and oci_sign_command already include `;` hence removing it from where those values are used.


maelvls:
Ouch, I pushed the tag v0.13.1 in the approver-policy repository. This bug prevented the release to go through and forced me to delete and recreate the tag ([details](https://github.com/cert-manager/approver-policy/pull/410/files))

PR #94, opened 2024-03-24
Upgrade cert-manager to v1.14.4 and start using cert-manager-startupapicheck instead of cert-manager-ctl

inteon:


Makefile
```diff
@@ -42,6 +42,9 @@ bin_dir := _bin
 $(bin_dir) $(bin_dir)/scratch:
 	mkdir -p $@
 
+# Set required variables
+build_names := #
```
inteon, line 46: small workaround required to make the root images-learn-sha target work.

PR #95, opened 2024-03-25
Add golangci-lint linting

inteon:
Adds golangci-lint with most linters enabled (I used `golangci-lint linters` to enable them and check whether their output is useful).
We can disable the warnings in each repo by using the following:
```yaml
issues:
  exclude-rules:
    - linters:
        - errcheck
        - staticcheck
        - gosimple
        - contextcheck
        - promlinter
        - prealloc
        - containedctx
        - gci
        - errorlint
        - gosec
        - unparam
        - gocritic
        - gofmt
      text: ".*"
```

### DIFF
--- a/modules/go/01_mod.mk
+++ b/modules/go/01_mod.mk
@@ -16,6 +16,12 @@ ifndef bin_dir
 $(error bin_dir is not set)
 endif
 
+ifndef repo_name
+$(error repo_name is not set)
+endif
+
+golangci_lint_override := $(dir $(lastword $(MAKEFILE_LIST)))/.golangci.override.yaml
+
 .PHONY: verify-govulncheck
 ## Verify all Go modules for vulnerabilities using govulncheck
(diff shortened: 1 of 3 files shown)


modules/go/01_mod.mk
```diff
@@ -39,3 +45,60 @@ verify-govulncheck: | $(NEEDS_GOVULNCHECK)
 				popd >/dev/null; \
 				echo ""; \
 			done
+
+ifdef golangci_lint_config
```
SgtCoDFish, line 49: review note: This is the most important part of the PR. golangci-lint won't be enabled unilaterally - it'll only be enabled if we do another step after merging the makefile-modules upgrade.

SgtCoDFish (approved):
/lgtm
/approve

I'm happy this won't break all PRs on all repos, so happy to merge this! Thanks 🚀

PR #103, opened 2024-04-04
fix: use yq instead of sed for go module

ThatsMrTalbot:


inteon (commented):
/approve
/lgtm

PR #104, opened 2024-04-09
feat: add module to create/publish olm bundle

ThatsMrTalbot:
### Reason for change

Adds a module to generate OLM bundle using the existing helm module. The helm chart is templated and ran through `operator-sdk generate bundle`. 

The generated bundle is then ran through a tool to produce an oci directory in the format the oci-publish module expects. 

We could use docker build to produce the image using the dockerfile `operator-sdk` generates, but that requires the docker runtime and a "docker buildx builder" that supports producing oci layout directories. By writing a small tool we remove the requirement for a container runtime.

We can re-use the existing code for oci publishing to publish the olm bundle.

### Changes

- Add `operator-sdk` to the tools module
- Update `oci-publish` module to allow us to add push targets in addition to what the `oci-build` module produces
- Add `olm-bundle` module that creates the oci directory and adds it as a target for the `oci-publish` module


inteon:
@wallrj FYI

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,52 @@
...
+# alm-examples annotation in the clusterserviceversion
+olm_example_directory ?= deploy/examples
+
+# Base directory for OLM bundle
+olm_bundle_base := bundle
+olm_manifests_base := $(olm_bundle_base)/manifests
+olm_metadata_base := $(olm_bundle_base)/metadata
+olm_scorecard_base := $(olm_bundle_base)/tests/scorecard
+
+# We want to allow both the .yml and .yaml extension, so we use wildcard to find
```
inteon, line 27: @ThatsMrTalbot why do we support the .yml extension also?

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,52 @@
...
+
+# Add OLM as a push target to the oci-publish module
+ifdef oci_olm_image_name
+push_names += olm
+endif
+
+oci_digest_path_olm := $(bin_dir)/scratch/image/oci-layout-olm.$(oci_olm_image_tag).digests
+oci_layout_path_olm := $(bin_dir)/scratch/image/oci-layout-olm.$(oci_olm_image_tag)
+
+generate-olm-project: | $$(NEEDS_YQ)
```
inteon, line 47:
```suggestion
.PHONY: generate-olm-project
generate-olm-project: | $$(NEEDS_YQ)
```

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,52 @@
...
+first_existing_file = $(firstword $(wildcard $1))
+first_existing_file_or_default = $(firstword $(wildcard $1) $(firstword $1))
+
+# Utility variables
+current_makefile := $(lastword $(MAKEFILE_LIST))
+current_makefile_directory := $(dir $(current_makefile))
+olm_base_dir := $(current_makefile_directory)/base
+
+# Default config
+olm_publish_repos ?=
```
inteon, line 12: Could you add comments to explain these variables?

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,52 @@
...
+
+# Default config
+olm_publish_repos ?= 
+olm_project_id ?=
+oci_olm_image_tag ?= $(VERSION)
+preflight_container_project_ids ?=
+
+# The example directory is used to find CRD examples, this is used to create the
+# alm-examples annotation in the clusterserviceversion
+olm_example_directory ?= deploy/examples
```
inteon, line 19: I wouldn't place the CRD examples in the deploy folder, since it relates to how to use the controller (instead of how to deploy it).
Maybe, we can also rename the folder to `cr_examples` for clarity.

modules/olm-bundle/00_mod.mk
inteon: In modules/olm-bundle/base/bundle/README.md you explicitly state that the files have to be `$(deploy_name).clusterserviceversion.yaml` `$(olm_metadata_base)/annotation.yaml` etc.
That will also simplify this part of the makefile a bit.

ThatsMrTalbot: Updated to just use yaml extensions

modules/olm-bundle/00_mod.mk
ThatsMrTalbot: I only chose this dir because it kind of follows the pattern that kubebuilder has, it puts its examples in `config/samples` where `config` mainly contains the manifests for deployment.

I could move it into the bundle directory, `bundle/examples`

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,52 @@
...
+ifdef oci_olm_image_name
+push_names += olm
+endif
+
+oci_digest_path_olm := $(bin_dir)/scratch/image/oci-layout-olm.$(oci_olm_image_tag).digests
+oci_layout_path_olm := $(bin_dir)/scratch/image/oci-layout-olm.$(oci_olm_image_tag)
+
+generate-olm-project: | $$(NEEDS_YQ)
+	cp -r $(olm_base_dir)/. ./
+	$(YQ) -i  '.projectName = "$(deploy_name)"' PROJECT
```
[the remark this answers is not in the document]
ThatsMrTalbot: I could create a olm_project_name variable and use that everywhere. I only used this one because in all the repos I looked at it contained a sensible value

modules/olm-bundle/00_mod.mk
ThatsMrTalbot: Added comments to all the config vars

modules/olm-bundle/00_mod.mk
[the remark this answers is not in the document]
inteon: Please create that `olm_project_name` variable, we can set it's value equal to the `deploy_name` value and explain that we use it because it is a reasonable value to use.
This will make it clear that there is no dependency on any deployment anywhere.

modules/olm-bundle/00_mod.mk
inteon: That would make sense, we also have eg. https://github.com/cert-manager/trust-manager/tree/main/docs/api docs which explain the API. We could maybe place those two together in the same dir?
Maybe `docs/examples`?

ThatsMrTalbot: I have moved it under bundle/examples and added a line to the bundle readme

modules/olm-bundle/00_mod.mk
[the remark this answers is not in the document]
ThatsMrTalbot: I have created olm_project_name and if deploy_name is set it uses that as a default.

modules/olm-bundle/00_mod.mk
```diff
@@ -0,0 +1,81 @@
+# Utility functions
+fatal_if_undefined = $(if $(findstring undefined,$(origin $1)),$(error $1 is not set))
+
+# Utility variables
+current_makefile := $(lastword $(MAKEFILE_LIST))
+current_makefile_directory := $(dir $(current_makefile))
+olm_base_dir := $(current_makefile_directory)/base
+
+ifdef build_name
```
inteon, line 9: @ThatsMrTalbot Why is this conditional on `build_name`? What is `build_name`?

modules/olm-bundle/01_mod.mk
```diff
@@ -0,0 +1,125 @@
...
+olm-bundle: $(helm_chart_archive) $(olm_clusterserviceversion_path) | $(NEEDS_HELM) $(NEEDS_YQ) $(NEEDS_OPERATOR-SDK) $(NEEDS_YAMLFMT)
+	rm -rf $(olm_bundle_dir) $(oci_layout_path_olm) $(oci_digest_path_olm)
+	mkdir -p $(olm_bundle_dir)
+	cd $(olm_bundle_dir) && $(HELM) template $(addprefix --values=,$(abspath $(olm_helm_value_path))) $(olm_project_name) $(CURDIR)/$(helm_chart_archive) |\
+		$(YQ) 'del(.metadata.annotations["helm.sh/resource-policy"])' |\
+		$(YQ) 'del(.metadata.labels["app.kubernetes.io/managed-by"])' |\
+		$(YQ) 'del(.metadata.labels["app.kubernetes.io/version"])' |\
+		$(YQ) 'del(.metadata.labels["helm.sh/chart"])' |\
+		$(YQ) 'del(.metadata.creationTimestamp)' |\
+		$(YQ) - $(abspath $(olm_clusterserviceversion_path)) $(abspath $(olm_examples)) $(abspath $(olm_additional_manifests)) |\
```
inteon, line 39: How does this work? Are the examples deployed by default when you install the olm bundle?

modules/olm-bundle/01_mod.mk
```diff
@@ -0,0 +1,125 @@
...
+		$(OPERATOR-SDK) generate bundle --output-dir . --package $(olm_project_name) --version $(VERSION:v%=%)
+	
+	@# Set the container image annodation
```
inteon, line 42:
```suggestion
	@# Set the container image annotation
```

modules/olm-bundle/01_mod.mk
```diff
@@ -0,0 +1,125 @@
...
+	$(YQ) -i '.metadata.annotations.containerImage = .spec.install.spec.deployments[0].spec.template.spec.containers[0].image' $(olm_bundle_dir)/manifests/$(olm_project_name).clusterserviceversion.yaml 
+	$(YQ) -i '.spec.relatedImages = [ .spec.install.spec.deployments[].spec.template.spec.containers[] | {"name": .name, "image": .image} ]' $(olm_bundle_dir)/manifests/$(olm_project_name).clusterserviceversion.yaml 
+
+	@# Set the supported OS/Arches based on $(oci_platforms)
+	$(YQ) -i --string-interpolation '. * ("$(oci_platforms)" | [ split(",").[] | split("/") | ["operatorframework.io/os.\(.[0])", "operatorframework.io/arch.\(.[1])"] ] | [ .[][] ] | unique | .[] as $$item ireduce({}; . * {"metadata": {"annotations": {$$item: "supported"}}}))' $(olm_bundle_dir)/manifests/$(olm_project_name).clusterserviceversion.yaml 
+	
+	@# Set the openshift version
+	$(YQ) -i '.annotations."com.redhat.openshift.versions"="v4.6"' $(olm_bundle_dir)/metadata/annotations.yaml
```
inteon, line 50: Do we place this version in a variable with a comment to explain what v4.6 means and when we should update it?

ThatsMrTalbot: I have done this now

modules/olm-bundle/00_mod.mk
ThatsMrTalbot: That is a copy paste error, I have fixed

inteon:
Thanks @ThatsMrTalbot !
/approve
/lgtm

PR #105, opened 2024-04-12
feat: add license module

ThatsMrTalbot:
# Reason for change

This PR adds a new `licenses` module, this module will:

- Add a generate target to generate a LICENCES file next to each go.mod 
- Add a layer to images built by oci-build that includes the licenses under the "/licenses" directory.

# Additional changes

As part of adding the functionality to add additional layers to the image, the image_tool was also re-worked. It now has less dependencies and uses a file per command.


modules/oci-build/01_mod.mk
```diff
@@ -64,6 +64,10 @@ $(oci_build_targets): oci-build-%: ko-config-% | $(NEEDS_KO) $(NEEDS_GO) $(NEEDS
 		--push=false \
 		--bare
 
+	$(IMAGE_TOOL) append-layers \
```
inteon, line 67: Does this break any sboms that we produce/ would produce in the future? I think it changes the digest, right?

ThatsMrTalbot: Our current solution (writing an OCI layout dir and pushing that) does not push sboms. It just writes them to a local file. We could use `cosign attest` to push these files in the future, this code does not break that.

modules/oci-build/image_tool/append_layers.go
```diff
@@ -0,0 +1,155 @@
...
+	var layer v1.Layer
+	if stat.IsDir() {
+		var buf bytes.Buffer
+
+		tw := tar.NewWriter(&buf)
+
+		filepath.Walk(path, func(target string, info fs.FileInfo, err error) error {
+			must("walk error", err)
+
+			header, err := tar.FileInfoHeader(info, info.Name())
```
inteon, line 54: @ThatsMrTalbot should we maybe overwrite the file owner and file ModTime, AccessTime, ... here?
So this becomes a reproducible build?
We also have the SOURCE_DATE_EPOCH environment variable that is used with ko.

inteon: I think we should also write some tests for this tool (maybe not as part of this PR).

inteon:
I just found one extra complication: https://github.com/cert-manager/makefile-modules/pull/105#discussion_r1565576600
No remarks for the rest of the PR.
Thanks for this feature, seems very promising!

modules/oci-build/image_tool/append_layers.go
ThatsMrTalbot: Good point about r reproducible builds. I have rewrote it to write a simplified tar header that removes the time information. 

This mimics ko's behaviour that does not seem to set modtime: https://github.com/ko-build/ko/blob/e407275248762120557098982154132dfec9a01f/pkg/build/gobuild.go#L530-L538

The SOURCE_DATE_EPOCH var sets the build time of the container, since this is a container config attribute this append tool does not actually change it from what ko sets.

modules/oci-build/image_tool/append_layers.go
```diff
@@ -0,0 +1,164 @@
...
+			must("could not build relative path", err)
+
+			// Write simplified header, this removes all fields that would cause
+			// the build to be non-reproducible (like modtime for example)
+			err = tw.WriteHeader(&tar.Header{
+				Typeflag: header.Typeflag,
+				Name:     name,
+				Mode:     header.Mode,
+				Linkname: header.Linkname,
+				Size:     header.Size,
```
inteon, line 67: What does the UID and GID default to?
When you ls in the container image, what is the result that you see?

inteon: ko also does not seem to set this value, so it is probably fine? https://github.com/ko-build/ko/blob/e407275248762120557098982154132dfec9a01f/pkg/build/gobuild.go#L635-L644
Maybe we should also overwrite the Mode like ko to 0555?

ThatsMrTalbot: 0555 includes execute permissions, which may be desirable for executables, but may not be for other files. 

UID and GID default to root (same as the ko binary). This is probably fine as the container runs with the `nobody` user, so having the files owned by root means they cant be modified.

inteon:
/approve
/lgtm
Thanks for the new capability!

PR #115, opened 2024-04-24
Make generate and verify work in parallel

inteon:
`make -j verify`. `make -j generate`, `make -j vendor-go verify` and `make -j vendor-go generate` should now all work as expected.

This was not an easy task. We use recursive $(MAKE) invocations in our makefiles and this makes things more complicated.
If a recursive $(MAKE) call requires tools to be installed, these tools will now be downloadable by this invocation without conflicting with other $(MAKE) invocations thanks to the file locking mechanism introduced in this PR.

TODO: not sure if flock works in darwin, we could just disable it on darwin and only enable it on linux and CI. This will require darwin users to run the commands in serial or download all tools beforehand.


modules/generate-verify/02_mod.mk
```diff
@@ -16,18 +16,24 @@
 ## Generate all generate targets.
 ## @category [shared] Generate/ Verify
 generate: $$(shared_generate_targets)
+	@echo "The following targets cannot be run simultaniously with each other or other generate scripts:"
+	$(foreach TARGET,$(shared_generate_targets_dirty), $(MAKE) $(TARGET))
```
ThatsMrTalbot, line 20: Does `shared_generate_targets_dirty` need a default empty value in 00_mod.mk?

modules/tools/util/lock.sh
```diff
@@ -0,0 +1,64 @@
...
+lockfile="$finalfile.lock"
+# Timeout in seconds.
+timeout=60
+
+mkdir -p "$(dirname "$lockfile")"
+touch "$lockfile"
+exec {FD}<>"$lockfile"
+
+# wait for the file to be unlocked
+if ! flock -x -w $timeout $FD; then
```
ThatsMrTalbot, line 37: flock does not seem to be installed on OSX, meaning I think this will break OSX by default

modules/generate-verify/02_mod.mk
inteon: Yes, that makes sense.

modules/tools/util/lock.sh
inteon: Ok, I'll just skip the locking on OSX then.

PR #116, opened 2024-04-24
feat: add ability to rewrite image name before pushing

ThatsMrTalbot:
There are two main changes:

- Previously we used `build_names` as a source for push targets, by using a different variable `push_names` we can create push targets for things that are not built with `oci-build` (which just builds golang containers)
- Allows specifying `image_registry_rewrite` rules in the format `vanity-domain.com=domain-we-push-to.com`. This allows us to know the final image names, as well as still be able to push to the correct registry.

### DIFF
--- a/modules/oci-publish/00_mod.mk
+++ b/modules/oci-publish/00_mod.mk
@@ -1,0 +1,44 @@
+# Push names is equivilent to build_names, additional names can be added for 
+# pushing images that are not build with the oci-build module
+push_names ?=
+push_names += $(build_names)
+
+# Sometimes we need to push to one registry, but pull from another. This allows
+# that.
+#
+# The lines should be in the format a=b
+#
+# The value on the left is the domain you include in your oci_<name>_image_name
+# variable, the one on the right is the domain that is actually pushed to.
+#
+# For example, if we set up a vanity domain for the current quay:
+# 
+#   oci_controller_image_name = registry.cert-manager.io/cert-manager-controller` 
+#   image_registry_rewrite += registry.cert-manager.io=quay.io/jetstack
+#
+# This would push to quay.io/jetstack/cert-manager-controller.
+#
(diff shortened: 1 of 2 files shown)


modules/oci-publish/00_mod.mk
```diff
@@ -0,0 +1,29 @@
...
+image_registry_rewrite ?= 
+
+# Utilities for extracting the key and value from a foo=bar style line
+kv_key = $(word 1,$(subst =, ,$1))
+kv_value = $(word 2,$(subst =, ,$1))
+
+# Apply the image_registry_rewrite rules, if no rules match an image then the 
+# image name is not changed. Any rules that match will be applied.
+#
+# For example, if there was a rule vanity-domain.com=real-registry.com/foo
```
inteon, line 17: Could you clarify here which of the two domains is the one that eg. would be used in your Helm chart?
Based on the description, it is not yet clear to me which is the registry for pushing and which is the one for writing.

ThatsMrTalbot: The one on the left is the domain you include in your `oci_<name>_image_name` variable, the one on the right is the domain that is actually pushed to

For example, if we set up a vanity domain for the current quay:

```
oci_controller_image_name = registry.cert-manager.io/cert-manager-controller` 
image_registry_rewrite += registry.cert-manager.io=quay.io/jetstack
```

This would push to `quay.io/jetstack/cert-manager-controller`.

The general idea is `oci_<name>_image_name` contains the _final_ image name, after replication, after vanity domains etc.

inteon: Ok, that makes a lot of sense, could you also put this example and info in the comments in this file?

ThatsMrTalbot: I have copied what I wrote into a comment in the code

inteon (commented):
/approve
/lgtm

PR #118, opened 2024-04-25
fix: use second expansion for dependencies in license module

ThatsMrTalbot:
Since licenses is before tools alphabetically the `$(NEEDS....)` vars will not exist yet in 00_mod.mk. This uses second expansion to fix dependencies


SgtCoDFish (approved):
/lgtm
/approve
/hold

Holding for a question: Is this not a use case for something like 02_mod.mk? i.e. would it work - and be simpler - to move the target to 02_mod.mk, or am I missing something?

SgtCoDFish:
I think it'd be cleaner to split it into 01_mod.mk personally but it's not a blocker, I'll unhold!

/unhold

PR #119, opened 2024-04-25
fix: make verify fails due to characters in generate target

ThatsMrTalbot:
Previously the generate target was a path, since the verify target prefixes this the new verify target name was invalid. This fixes that


SgtCoDFish (approved):
/lgtm
/approve

modules/licenses/00_mod.mk
```diff
@@ -24,7 +24,7 @@ $(generate_go_licenses_targets): | $(NEEDS_GO-LICENSES)
 ## Generate licenses for the golang dependencies
 ## @category [shared] Generate/Verify
 generate-go-licences: $(generate_go_licenses_targets)
-shared_generate_targets += $(generate_go_licenses_targets)
+shared_generate_targets += generate-go-licences
```
inteon, line 27:
```suggestion
shared_generate_targets += generate-go-licenses
```

PR #120, opened 2024-04-25
fix: image rewriting not working on push targets

ThatsMrTalbot:


SgtCoDFish (approved):
/lgtm
/approve

Waving through untested!

PR #121, opened 2024-04-25
Small bug fixes and improvements

inteon:
Fixes some issues missed in #115.
Also improves some other aspects (eg. prevents an incorrect warning from printing).

### DIFF
--- a/modules/generate-verify/util/verify.sh
+++ b/modules/generate-verify/util/verify.sh
@@ -44,8 +44,7 @@ cleanup() {
 }
 trap "cleanup" EXIT SIGINT
 
-cp -a "${projectdir}/." "${tmp}"
-rm -rf "${tmp}/_bin" # clear all cached files
+rsync -aEq "${projectdir}/." "${tmp}" --exclude "_bin/"
 pushd "${tmp}" >/dev/null
 
 "$@"
--- a/modules/help/help.sh
+++ b/modules/help/help.sh
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu -o pipefail
+set -o errexit
+set -o nounset
+set -o pipefail
 
 ## 1. Build set of extracted line items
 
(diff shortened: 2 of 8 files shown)


modules/generate-verify/util/verify.sh
```diff
@@ -44,8 +44,7 @@ cleanup() {
 }
 trap "cleanup" EXIT SIGINT
 
-cp -a "${projectdir}/." "${tmp}"
-rm -rf "${tmp}/_bin" # clear all cached files
+rsync -aEq "${projectdir}/." "${tmp}" --exclude "_bin/"
```
inteon, line 47: The `cp` command was throwing errors because the `_bin` folder was being changed while the `cp` was running.
Instead, we can use rsync, which has an exclude option for the `_bin/` folder.

modules/help/help.sh
```diff
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu -o pipefail
+set -o errexit
+set -o nounset
+set -o pipefail
```
inteon, line 19: I updated all scripts to have consistent set headers everywhere.

modules/tools/00_mod.mk
```diff
@@ -576,6 +576,7 @@ $(DOWNLOAD_DIR)/tools/rclone@$(RCLONE_VERSION)_$(HOST_OS)_$(HOST_ARCH): | $(DOWN
 MISSING=$(shell (command -v curl >/dev/null || echo curl) \
              && (command -v sha256sum >/dev/null || command -v shasum >/dev/null || echo sha256sum) \
              && (command -v git >/dev/null || echo git) \
+             && (command -v rsync >/dev/null || echo rsync) \
```
inteon, line 579: We now use rsync in the `verify.sh` script.

modules/tools/util/lock.sh
```diff
@@ -66,6 +68,5 @@ finish() {
     rm -rf "$finalfile" || true
   fi
   rm -rf "$lockfile" || true
-  exit $rv
```
inteon, line 69: This exit was not needed, trap keeps the exit code in case there was an error.

modules/tools/util/lock.sh
```diff
@@ -66,6 +68,5 @@ finish() {
...
 }
-trap finish EXIT
+trap finish EXIT SIGINT
```
inteon, line 72: I added SIGINT to be consistent with `verify.sh`.

PR #122, opened 2024-04-26
fix: use official preflight binaries when available

ThatsMrTalbot:
You can only submit preflight check results using an offical binary, it submits the version and build commit as part of the submission and this is validated by RedHat

This change uses the offical binaries when available so we can submit test results


modules/tools/00_mod.mk
```diff
@@ -560,6 +559,23 @@ $(DOWNLOAD_DIR)/tools/rclone@$(RCLONE_VERSION)_$(HOST_OS)_$(HOST_ARCH): | $(DOWN
...
+PREFLIGHT_linux_arm64_SHA256SUM=c42cf4132027d937da88da07760e8fd9b1a8836f9c7795a1b60513d99c6939fe
+
+# Currently there are no offical releases for darwin, you cannot submit results 
+# on non-official binaries, but we can still run tests.
+#
+# Once https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/942 is merged
+# we can remove this darwin specific hack
+$(bin_dir)/downloaded/tools/preflight@$(PREFLIGHT_VERSION)_darwin_%: | $(bin_dir)/downloaded/tools
+	GOWORK=off GOBIN=$(CURDIR)/$(dir $@) $(GO) install github.com/redhat-openshift-ecosystem/openshift-preflight/cmd/preflight@$(PREFLIGHT_VERSION)
+	@mv $(CURDIR)/$(dir $$@)/preflight $@
```
inteon, line 572: @ThatsMrTalbot You now have to use the `@source $(lock_script) $@; \` logic to make sure these tools can be downloaded concurrently. Also make sure to write the result to the `$(outfile)`.

inteon (commented):
Awesome!
/approve
/lgtm

PR #124, opened 2024-04-26
Use ln instead of cp to fix issues with concurrent tool setup

inteon:
There is still one type of issue that we see after implementing the locking mechanism for tool downloads: https://storage.googleapis.com/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_cert-manager/6963/pull-cert-manager-master-make-verify/1783870541019484160/build-log.txt

This error happens when two Make processes are downloading tools in the same folder (not eg. a copy of the project as is the case for the verify commands).
The files in the DOWNLOAD_DIR are downloaded cleanly as expected. However, the "link" to $(bin_dir)/tools/xxxx is created by copying the file from the DOWNLOAD_DIR to this new location. This copy action is also prone to errors when performed concurrently. If, instead, we just call `ln -fs`, this operation is atomic and we fix this issue.
The reason that we used `cp` instead of `ln -fs` is explained in the old comment and no longer applies, the $(bin_dir)/downloaded folder is no longer mounted directly in the pod.

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -167,22 +167,11 @@ $(bin_dir)/scratch/%_VERSION: FORCE | $(bin_dir)/scratch
 # --retry-connrefused = retry even if the initial connection was refused
 CURL = curl --silent --show-error --fail --location --retry 10 --retry-connrefused
 
-# In Prow, the pod has the folder "$(bin_dir)/downloaded" mounted into the
-# container. For some reason, even though the permissions are correct,
-# binaries that are mounted with hostPath can't be executed. When in CI, we
-# copy the binaries to work around that. Using $(LN) is only required when
-# dealing with binaries. Other files and folders can be symlinked.
-#
-# Details on how "$(bin_dir)/downloaded" gets cached are available in the
-# description of the PR https://github.com/jetstack/testing/pull/651.
-#
-# We use "printenv CI" instead of just "ifeq ($(CI),)" because otherwise we
-# would get "warning: undefined variable 'CI'".
-ifeq ($(shell printenv CI),)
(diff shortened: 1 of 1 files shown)


modules/tools/00_mod.mk
```diff
@@ -167,22 +167,11 @@ $(bin_dir)/scratch/%_VERSION: FORCE | $(bin_dir)/scratch
 # --retry-connrefused = retry even if the initial connection was refused
 CURL = curl --silent --show-error --fail --location --retry 10 --retry-connrefused
 
-# In Prow, the pod has the folder "$(bin_dir)/downloaded" mounted into the
-# container. For some reason, even though the permissions are correct,
```
inteon, line 171: This is no longer the case: before the test, files are copied into this folder and after the test, files are copied out of this folder.

modules/tools/00_mod.mk
```diff
@@ -204,8 +193,8 @@ TOOL_NAMES :=
 #        in targets or in scripts, because it is agnostic to the
 #        working directory
 # - an unversioned target $(bin_dir)/tools/xxx is generated that
-#   creates a copy/ link to the corresponding versioned target:
-#   $(bin_dir)/tools/xxx@$(XXX_VERSION)_$(HOST_OS)_$(HOST_ARCH)
+#   creates a link to the corresponding versioned target:
+#   $(DOWNLOAD_DIR)/tools/xxx@$(XXX_VERSION)_$(HOST_OS)_$(HOST_ARCH)
```
maelvls, line 197: FWIW this change broke some users of the `ADDITIONAL_TOOLS` API: https://github.com/cert-manager/website/pull/1644#issue-2845434199

PR #127, opened 2024-04-29
fix: allow flock timeout to be configurable

ThatsMrTalbot:


SgtCoDFish (approved):
/lgtm
/approve

Makes sense to me!

PR #132, opened 2024-05-01
Fix find invocations on macOS

### LINKED ISSUE #130: Broken use of `find` on macOS
In a couple of places (e.g. [here](https://github.com/cert-manager/makefile-modules/blob/f053c2c92459763f424280dde8190aae674743b9/modules/go/01_mod.mk#L49) but there are others) we use the `-printf` argument to find.

That isn't available on macOS so those commands fail:

```console
find: -printf: unknown primary or operator
```

It's not possible to run `make verify-govulncheck` or `make verify-golangci-lint` locally on macOS

SgtCoDFish:
Closes #130 

1. Removes non-portable arg for `find`
2. Vendors gci, which was previously assumed to be installed on the system
3. Removes some trailing whitespace from the tools module


inteon (commented):
Works on my laptop
/approve
/lgtm

PR #148, opened 2024-05-20
fix: ensure we set the correct media type on appended oci layers

ThatsMrTalbot:
This is a subtle bug.

The index can be OCI while the image manifest is in docker format.

So we need to set the layer media type on an image by image basis.

### DIFF
--- a/modules/oci-build/image_tool/append_layers.go
+++ b/modules/oci-build/image_tool/append_layers.go
@@ -52,31 +52,49 @@ var CommandAppendLayers = cobra.Command{
 		index, err := path.ImageIndex()
 		must("could not load oci image index", err)
 
-		indexMediaType, err := index.MediaType()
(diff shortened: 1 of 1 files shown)


modules/oci-build/image_tool/append_layers.go
```diff
@@ -52,31 +52,53 @@ var CommandAppendLayers = cobra.Command{
...
+			return img
+		})
 
 		_, err = layout.Write(oci, index)
 		must("could not write image", err)
 	},
 }
 
-func loadLayerFromDirOrTarball(path string, mediaType types.MediaType) v1.Layer {
+type UntypedLayer struct {
```
inteon, line 85: @ThatsMrTalbot can we keep all structs and functions private?

ThatsMrTalbot: Updated this now

PR #149, opened 2024-05-20
fix: pushing olm bundles to github was failing due to missing credentials

ThatsMrTalbot:
When using the github cli to clone it uses `-c "credential.helper=!<path to self> auth git-credential"` on the git command to provide credentials for the clone. This means that the `gh repo clone` works. 

However due to this credentials are never set up on the repo so the push will be denied.

The fix is to do the same thing the github cli does, and set the github CLI as a credential helper for the push command.

I also make the "pr create" command not fail the build as if the PR already exists its not an error state.


SgtCoDFish:
I think @wallrj would be good to look at this - I don't really have any context on it or any ability to evaluate whether this is a good change or not personally!

inteon (commented):
Looks good as far as I can see.
We might have to re-evaluate this logic once we migrate the OSS cert-manager OLM module to this logic.
/approve
/lgtm

PR #152, opened 2024-05-28
fix: bash wants a semi-colon before end braces or the brace is treated like an argument

ThatsMrTalbot:

### DIFF
--- a/modules/olm-bundle/01_mod.mk
+++ b/modules/olm-bundle/01_mod.mk
@@ -95,7 +95,7 @@ olm-publish-$(subst /,-,$1): olm-bundle | $(NEEDS_GH) $(bin_dir)/scratch
 		git add operators/$(olm_project_name)/ci.yaml && \
 		git commit -m "operator $(olm_project_name) ($(VERSION))" && \
 		git push -f -c "credential.helper=!$(GH) auth git-credential" origin $(VERSION) && \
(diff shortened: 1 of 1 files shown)


modules/olm-bundle/01_mod.mk
```diff
@@ -95,7 +95,7 @@ olm-publish-$(subst /,-,$1): olm-bundle | $(NEEDS_GH) $(bin_dir)/scratch
 		git add operators/$(olm_project_name)/ci.yaml && \
 		git commit -m "operator $(olm_project_name) ($(VERSION))" && \
 		git push -f -c "credential.helper=!$(GH) auth git-credential" origin $(VERSION) && \
-		{ $(GH) pr create --repo $1 --head $(firstword $(subst /, ,$2)):$(VERSION) --title "operator $(olm_project_name) ($(VERSION))" ||: }
+		{ $(GH) pr create --repo $1 --head $(firstword $(subst /, ,$2)):$(VERSION) --title "operator $(olm_project_name) ($(VERSION))" ||:; }
```
inteon, line 98: @ThatsMrTalbot could you explain why we need to put this last line between brackets?

ThatsMrTalbot: Its the `||:` at the end (which is just a shorthand for `|| true`.

I only want the `|| true` to apply if the `pr create` fails, as it exits 1 if the PR already exists.

Without the braces the `|| true` applies to all the commands, with the braces its just applied to the one.

inteon: Ah thanks, I did not know about the `||:` shorthand.

PR #171, opened 2024-07-17
Update kind.mk

jabdoa2:
Commit change from https://github.com/cert-manager/trust-manager/pull/387 in the correct location

### DIFF
--- a/modules/kind/kind.mk
+++ b/modules/kind/kind.mk
@@ -55,9 +55,16 @@ $(kind_kubeconfig): $(kind_cluster_config) $(bin_dir)/scratch/cluster-check | im
 	$(KIND) get kubeconfig --name $(kind_cluster_name) > $@
 
 .PHONY: kind-cluster
+kind-cluster: $(kind_kubeconfig)
+
+.PHONY: kind-cluster-load
 ## Create Kind cluster and wait for nodes to be ready
+## Load the kubeconfig into the default location so that
+## it can be easily queried by kubectl. This target is
+## meant to be used directly, NOT as a dependency.
+## Use `kind-cluster` as a dependency instead.
 ## @category [shared] Kind cluster
-kind-cluster: $(kind_kubeconfig) | $(NEEDS_KUBECTL)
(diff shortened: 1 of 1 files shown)


inteon:
I updated the fix based on how I think this target was originally designed.
It looks like `kind-cluster` was not meant to be used as a dependency by other targets, instead they should have depended on `$(kind_kubeconfig)`.
This PR replaces `kind-cluster` with a simple "depend on `$(kind_kubeconfig)`" target and renames the old `kind-cluster` target to `kind-cluster-load`.

PTAL and let me know what you think.

jabdoa2:
> I updated the fix based on how I think this target was originally designed. It looks like `kind-cluster` was not meant to be used as a dependency by other targets, instead they should have depended on `$(kind_kubeconfig)`. This PR replaces `kind-cluster` with a simple "depend on `$(kind_kubeconfig)`" target and renames the old `kind-cluster` target to `kind-cluster-load`.
> 
> PTAL and let me know what you think.

Not sure. Even in that case it should not silently overwrite my kubectl config and purge all my cluster configs. At least ask with a huge warning. Took me a few hour to get all my clusters configured again. Ideally it would merge into an existing config or we could use a separate config since we set the environment variable anyway.

inteon:
> > I updated the fix based on how I think this target was originally designed. It looks like `kind-cluster` was not meant to be used as a dependency by other targets, instead they should have depended on `$(kind_kubeconfig)`. This PR replaces `kind-cluster` with a simple "depend on `$(kind_kubeconfig)`" target and renames the old `kind-cluster` target to `kind-cluster-load`.
> > PTAL and let me know what you think.
> 
> Not sure. Even in that case it should not silently overwrite my kubectl config and purge all my cluster configs. At least ask with a huge warning. Took me a few hour to get all my clusters configured again. Ideally it would merge into an existing config or we could use a separate config since we set the environment variable anyway.

It is surprising that the kubectl config is purged, normally the command should merge the kubeconfig into your global kubeconfig. It seems like there is another bug here that we should fix.

inteon (commented):
Let's start with this change and gather some feedback.
/approve
/lgtm

PR #176, opened 2024-07-29
fix: auto-upgrade not working for OLM bundles

ThatsMrTalbot:
Currently the PRs generated for OLM publishing do not auto-upgrade. This is for two reasons:
- The way upgrades are managed defaults to "replaces-mode", in which each version specifies the version it is replacing
- We do not specify the version we are replacing

This PR changes it so we default to "semver-mode" where the semver of the release is used to determine the version we are replacing

It also specifies the replaces field on top of this as the certified-operators repo seems to only use the "replaces-mode"

This should handle all cases.

### DIFF
--- a/modules/olm-bundle/00_mod.mk
+++ b/modules/olm-bundle/00_mod.mk
@@ -45,6 +45,9 @@ olm_publish_repos ?=
 # program
 olm_project_id ?=
 
+# Used for handling auto-upgrade, the version that this release is replacing
+olm_replaces_version ?= $(shell git describe --tags --always --match='v*' --abbrev=0 --exclude $(VERSION))
+
 # OLMs are built as images, so can be pushed to image registries. We default the
 # tag to the version.
 #
--- a/modules/olm-bundle/01_mod.mk
+++ b/modules/olm-bundle/01_mod.mk
@@ -53,6 +53,9 @@ olm-bundle: $(helm_chart_archive) $(olm_clusterserviceversion_path) | $(NEEDS_HE
 		$(YQ) - $(abspath $(olm_clusterserviceversion_path)) $(abspath $(olm_examples)) $(abspath $(olm_additional_manifests)) |\
 		$(OPERATOR-SDK) generate bundle --output-dir . --package $(olm_project_name) --version $(VERSION:v%=%)
 
+	@# Set the previous version in the replaces field, this is required for auto-upgrade to work
+	$(YQ) -i '.spec.replaces = "$(olm_project_name).$(olm_replaces_version)"' $(olm_bundle_dir)/manifests/$(olm_project_name).clusterserviceversion.yaml
+
 	@# Set the container image annotation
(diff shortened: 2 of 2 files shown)


modules/olm-bundle/01_mod.mk
```diff
@@ -106,6 +109,7 @@ olm-publish-$(subst /,-,$1): olm-bundle | $(NEEDS_GH) $(bin_dir)/scratch
 		touch operators/$(olm_project_name)/ci.yaml && \
 		$(if $(and $(findstring redhat-openshift-ecosystem/certified-operators,$1),$(olm_project_id)), \
 			$(YQ) -i '.cert_project_id = "$(olm_project_id)"' operators/$(olm_project_name)/ci.yaml &&) \
+		$(YQ) -i '.updateGraph = "semver-mode"' operators/$(olm_project_name)/ci.yaml && \
```
wallrj, line 112: I had a vague memory that you couldn't change this setting, but the documentation says that this is supported, so perhaps  I'm misremembering or the limitation has since been lifted.
 * https://redhat-openshift-ecosystem.github.io/operator-pipelines/users/operator-ci-yaml/#operator-versioning

modules/olm-bundle/01_mod.mk
```diff
@@ -53,6 +53,9 @@ olm-bundle: $(helm_chart_archive) $(olm_clusterserviceversion_path) | $(NEEDS_HE
 		$(YQ) - $(abspath $(olm_clusterserviceversion_path)) $(abspath $(olm_examples)) $(abspath $(olm_additional_manifests)) |\
 		$(OPERATOR-SDK) generate bundle --output-dir . --package $(olm_project_name) --version $(VERSION:v%=%)
 
+	@# Set the previous version in the replaces field, this is required for auto-upgrade to work
+	$(YQ) -i '.spec.replaces = "$(olm_project_name).$(olm_replaces_version)"' $(olm_bundle_dir)/manifests/$(olm_project_name).clusterserviceversion.yaml
```
wallrj, line 57: Seems like this should work, based on the documentation.
 * https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#replaces

Maybe  `skipRange` would be worth adding in a future PR, to avoid having to upgrade through every patch version.

wallrj (approved):
I didn't test this but the changes make sense.

/lgtm
/approve

modules/olm-bundle/00_mod.mk
```diff
@@ -45,6 +45,9 @@ olm_publish_repos ?=
 # program
 olm_project_id ?=
 
+# Used for handline auto-upgrade, the version that this release is replacing
```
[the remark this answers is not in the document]
ThatsMrTalbot: Good catch, fixed

wallrj (approved):
/lgtm
/unhold

PR #181, opened 2024-08-16
fix: gci write should skip generated/vendor

erikgb:
Some generated files in trust-manger doesn't follow the gci-configured order of imports, which makes it harder to use `make fix-golangci-lint` in trust-manager. This PR configures gci to skip generated (and vendor) files. This aligns with the default gci configuration in golangci-lint: https://golangci-lint.run/usage/linters/#gci

I wonder if/why we need to run gci separately in `make fix-golangci-lint` - as the golangci-lint docs state that gci supports autofix, but that can eventually be addressed as a follow-up.

image


inteon:
Thank you for fixing this bug!
/approve
/lgtm

PR #190, opened 2024-09-10
Add govet shadow linter

inteon:
Add govet shadow linter.

FYI @erikgb


erikgb:
It will be interesting to see how many violations we have! 😅

inteon:
UPDATE: this causes a TON of violations, will create a PR to revert this change.

PR #191, opened 2024-09-16
Suggest applying the generated diffs

jsoref:
The output was expensive to calculate, and the diff output is machine generated. `patch` is fully capable of applying it, so let people know that if they are comfortable with diffs they are welcome to apply them.

Otherwise, people can still run the command to overwrite things if they have the cycles.


SgtCoDFish (approved):
/lgtm
/approve

Cheers, this is an easy one to merge :grin: Appreciate it!

PR #200, opened 2024-10-07
Add the -n flag to the 'ln' command

inteon:
This will make such that destination directories are overwritten if they exist already (instead of creating a new directory link inside the destination directory).

Before (when upgrading the vendored Go version):
1. `_bin/tools/goroot` points to go 1.23.1
2. `_bin/tools/goroot/goroot` was created and points to go 1.23.2

Now:
1. `_bin/tools/goroot` points to go 1.23.1
2. `_bin/tools/goroot` was replaced and points to go 1.23.2


SgtCoDFish (approved):
/lgtm
/approve

Cheers!

PR #214, opened 2024-11-26
Make generated CRDs directly available for testing and development

erikgb:
This is a new attempt at what was started in https://github.com/cert-manager/makefile-modules/pull/209, but accidentally got merged before being fully reviewed (and then reverted in https://github.com/cert-manager/makefile-modules/pull/213).

/hold
/cc @inteon 
/cc @ThatsMrTalbot


modules/helm/crds.mk
```diff
@@ -37,29 +37,29 @@ ifeq ($(HOST_OS),darwin)
 	sed_inplace := sed -i ''
 endif
 
+crds_dir ?= config/crd/bases
```
erikgb, line 40: I think it could make sense to have a proper default for this path. Alternatives:

1. deploy/crds ([cert-manager](https://github.com/cert-manager/cert-manager/tree/master/deploy/crds))
2. test/testdata (suggested in https://github.com/cert-manager/trust-manager/pull/482)

modules/helm/crds.mk
```diff
@@ -37,29 +37,29 @@ ifeq ($(HOST_OS),darwin)
...
 .PHONY: generate-crds
 ## Generate CRD manifests.
 ## @category [shared] Generate/ Verify
 generate-crds: | $(NEEDS_CONTROLLER-GEN) $(NEEDS_YQ)
-	$(eval crds_gen_temp := $(bin_dir)/scratch/crds)
 	$(eval directories := $(shell ls -d */ | grep -v -e 'make' $(shell git check-ignore -- * | sed 's/^/-e /')))
 
-	rm -rf $(crds_gen_temp)
-	mkdir -p $(crds_gen_temp)
+	mkdir -p $(crds_dir)
```
erikgb, line 48: I agree with comments on the previous PR that this no longer belongs in the helm makefile-module. But we could do this refactoring as a separate change? WDYT? My knowledge for makefile-modules is limited, so if I am going to do this refactoring, someone needs to show me how to declare makefile-module dependencies. Any examples already present?

modules/helm/crds.mk
SgtCoDFish: I actually quite like the idea of using `test/...`. One of the concerns around this was that we don't want people to think that these CRDs are a blessed thing they can use for deployments. In cert-manager we use deploy/crds because it works a bit differently, but for other projects I think `test/...` would signal to people that we don't necc. support them downloading these checked-in CRDs and using them.

My only suggestion there would be that for `testdata` I'd be expecting files, like [here](https://github.com/cert-manager/cert-manager/tree/e9ca5ee823952e2c8a83c2e3eddf0bb31c454fd2/internal/pem).

Maybe `test/crds`?

Ultimately it doesn't matter much though, I don't feel too strongly. My preferences in order would be:

1. `test/crds`
2. `test/testdata`
3. `deploy/crds`
4. Anywhere else

modules/helm/crds.mk
SgtCoDFish: I think we do really need more docs around makefile-modules

I'd be _fairly_ comfortable leaving this in the helm module with a TODO comment, but I worry that might actually make it more difficult to migrate later - it's unlikely, but if someone wanted to use an older version of the Helm module that supported these CRDs, they wouldn't be able to do so with the new crds module we'd create later.

I think for that reason I'm softly in favour of moving this to a different place in this PR, but it's not a blocker for me (and I don't currently have bandwidth to help with it so it's very easy for me to say that 😂 )

modules/helm/crds.mk
hawksight: Option 1 `deploy/crds` makes sense to me assuming this is where the RAW non-helm'd CRDs will go?

modules/helm/crds.mk
hawksight: I'm in favour of incremental improvements. I'd say keep it here, add a TODO / issue for the wider change.

modules/helm/crds.mk
erikgb: As @SgtCoDFish is stating, cert-manager is a bit special in this area. It is using schemapatch, and not generate controller-gen command, probably for "historical" reasons. So I would vote for staying away from deploy/crds per default.

Putting it under "test" feels a bit wrong to me, since these files will be used as input to the Helm charts. But absolutely no blocker for me. 😊

erikgb: Per https://github.com/cert-manager/trust-manager/pull/475#discussion_r1858035321, I think we should make room for also checking in generating webhook configuration resources (and RBAC roles and clusterroles).

SgtCoDFish: > Putting it under "test" feels a bit wrong to me, since these files will be used as input to the Helm charts. But absolutely no blocker for me. 😊

Ah, I'm not sure I'd realised that they would be used as inputs for the helm charts. In that case, something like `deploy/crds` starts to make more sense.

If we want to discourage use, something like `deploy/internal/crds`?

ThatsMrTalbot (changes requested):
Can you update the path to something like `deploy/crds` as discussed in the earlier thread, I also think we should drop a README in that dir with something like:

> cert-manager recommends the CRDs are deployed as part of the Helm chart, these CRDs exist for reference and development.

After this merges we can create an github issue to refactor it out the helm module.

erikgb:
> Can you update the path to something like `deploy/crds` as discussed in the earlier thread, I also think we should drop a README in that dir with something like:
> 
> 
> 
> > cert-manager recommends the CRDs are deployed as part of the Helm chart, these CRDs exist for reference and development.
> 
> 
> 
> After this merges we can create an github issue to refactor it out the helm module.

Will do! Thanks for reviewing @ThatsMrTalbot!

erikgb:
@ThatsMrTalbot, your initial review comments should be addressed now. I have also created a WIP PR to verify/test that this would actually work in our projects: https://github.com/cert-manager/trust-manager/pull/493. Rhere are probably additional (minor) details that need to be adjusted. PTAL!

modules/helm/crds.mk
erikgb: Seems like we agree to make this initial change smaller and improve afterwards.

modules/helm/crds.mk
erikgb: The default path is now to changed to `deploy/crds`.

modules/helm/crds.mk
```diff
@@ -37,29 +37,31 @@ ifeq ($(HOST_OS),darwin)
...
+	cat $(crds_dir_readme) > $(crds_dir)/README.md
```
SgtCoDFish, line 50: nitpick (non-blocking): I find the use of `cat` here a bit surprising - I'd think `cp` would be more appropriate? (`cat` has to read the file and output it - `cp` doesn't need to do that)

SgtCoDFish, line 50: nitpick: The makefile would be a little more efficient if `$(crds_dir)/README.md` was its own target, since its contents depend only on one file - we don't need to update the readme every time we run `generate-crds`, we only need to update the README if `crds_dir.README.md` has changed.

Pseudo code:

```make
$(crds_dir):
<tab> mkdir -p $@

$(crds_dir)/README.md: path/to/crds_dir.README.md | $(crds_dir)
<tab> cp $< $@

generate-crds: $(crds_dir)/README.md | $(NEEDS_CONTROLLER-GEN) $(NEEDS_YQ) $(crds_dir)
...
```

SgtCoDFish (commented):
A couple of little makefile things - what do you think? I don't feel super strongly but I think these might be worth updating unless I'm missing something!

modules/helm/crds.mk
erikgb: Great suggestion, Ashley! I have to understand a bit more about how the paths are set in makefile-modules, but this should be doable! Thanks!

erikgb: Agreed, and should be fixed now.

erikgb: I think I've got this working now. PTAL!

erikgb:
@ThatsMrTalbot @inteon Are you able to review this again? All remarks should be addressed now.

ThatsMrTalbot (approved):

SgtCoDFish:
/unhold

I reckon this is good to go!

inteon:
@erikgb it looks like the README is not auto-generated currently: https://github.com/cert-manager/approver-policy/pull/545

erikgb:
> @erikgb it looks like the README is not auto-generated currently: [cert-manager/approver-policy#545](https://github.com/cert-manager/approver-policy/pull/545)

Strange! I will take a look.

PR #216, opened 2024-12-05
Tools: bump openshift-preflight, gh, operator-sdk and others

maelvls:
While working on an internal Venafi project (namely, [venafi-installer-operator](https://gitlab.com/venafi/vaas/applications/tls-protect-for-k8s/venafi-installer-operator/-/merge_requests/67#note_2242999888)), I found that the openshift-preflight image tag we used wasn't accepted by Red Hat anymore:

```
time="2024-12-05T09:17:59Z" level=error msg=release-url error="invalid preflight version, please download the latest version and re-submit"
Error: could not submit to pyxis: could not create test results: status code: 400: body: {"type": "about:blank", "title": "Bad Request", "detail": "Validation error: 'github.com/redhat-openshift-ecosystem/openshift-preflight' version '1.10.0' is not supported. Supported versions are: ['1.10.1', '1.10.2']", "status": 400, "trace_id": "0xe9a0d036a9535370c17a24daf58bc8d7"}
```

I've also updated the other tools, but haven't updated `syft`. I've kept the old version (0.100.0, Jan 2024) because all of the latest versions use a replace statement, and thus cannot be installed using `go build`:

```
go: github.com/anchore/syft/cmd/syft@v1.17.0 (in github.com/anchore/syft@v1.17.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -116,28 +116,30 @@ tools += ginkgo=$(detected_ginkgo_version)
 tools += klone=v0.1.0
 # https://pkg.go.dev/github.com/goreleaser/goreleaser?tab=versions
 tools += goreleaser=v1.26.2
-# https://pkg.go.dev/github.com/anchore/syft/cmd/syft?tab=versions
+# https://pkg.go.dev/github.com/anchore/syft/cmd/syft?tab=versions. We are still
+# using an old version (0.100.0, Jan 2024) because all of the latest versions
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

(Haven't checked the version numbers but LGTM and happy to review a followup if needed)

PR #218, opened 2024-12-09
Remove helm_docs_use_helm_tool, helm_generate_schema and helm_verify_values options

inteon:
All repos using the Helm module now have these options enabled, making them useless and adding extra complexity.

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -73,8 +73,6 @@ $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_d
 ## @category [shared] Helm Chart
 helm-chart: $(helm_chart_archive)
 
-ifdef helm_docs_use_helm_tool
-
 helm_tool_header_search ?= ^<!-- AUTO-GENERATED -->
 helm_tool_footer_search ?= ^<!-- /AUTO-GENERATED -->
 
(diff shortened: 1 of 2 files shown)


modules/helm/helm.mk
```diff
@@ -83,35 +81,24 @@ helm_tool_footer_search ?= ^<!-- /AUTO-GENERATED -->
 ## @category [shared] Generate/ Verify
 generate-helm-docs: | $(NEEDS_HELM-TOOL)
 	$(HELM-TOOL) inject -i $(helm_chart_source_dir)/values.yaml -o $(helm_chart_source_dir)/README.md --header-search "$(helm_tool_header_search)" --footer-search "$(helm_tool_footer_search)"
-else
-.PHONY: generate-helm-docs
-## Generate Helm chart documentation.
-## @category [shared] Generate/ Verify
-generate-helm-docs: | $(NEEDS_HELM-DOCS)
-	$(HELM-DOCS) $(helm_chart_source_dir)/
```
erikgb, line 91: Could `NEEDS_HELM-DOCS` and/or `HELM-DOCS` be deleted, or is it still in use?

inteon: Indeed, I removed them here: https://github.com/cert-manager/makefile-modules/pull/218/files#diff-cd4027dc82a354f5f35bd8a151160a66163403b3436ccac88e05c555ffbbca78L98-L99

erikgb: Ok, thanks. Is `NEEDS_xyz` a built-in make feature, or some convention used here?

erikgb (commented):
/lgtm

modules/helm/helm.mk
inteon: It is a variable that we create based on the values in the tools variable.

PR #219, opened 2024-12-09
Helm chart OCI publish

inteon:
Adds `helm-chart-oci-push` target and enforce v-prefixed Helm chart versions as primary version.
Additionally, the `helm-chart-oci-push` target now publishes a non-v-prefixed version which improves Helm compatibility and provides backwards compatibility for some projects that only published a non-v-prefixed version before.

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -24,17 +24,15 @@ ifndef helm_chart_source_dir
 $(error helm_chart_source_dir is not set)
 endif
 
-ifndef helm_chart_name
-$(error helm_chart_name is not set)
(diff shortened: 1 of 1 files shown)


erikgb:
/lgtm

But make is non-trivial. 😉

wallrj (commented):
Suggest updating the [test-project](https://github.com/cert-manager/makefile-modules/tree/main/tests/e2e-projects/test-project) and adding some notes about these new targets to a README file in https://github.com/cert-manager/makefile-modules/tree/main/modules/helm

Explain the interplay between these new OCI related targets and those in the oci-build and oci-publish modules.

PR #220, opened 2024-12-10
fix: should include README.md in generated CRDs directory

erikgb:
This is a bugfix to https://github.com/cert-manager/makefile-modules/pull/214, see https://github.com/cert-manager/makefile-modules/pull/214#issuecomment-2532266310.

/cc @inteon


inteon (commented):
/approve
/lgtm

PR #221, opened 2024-12-11
Fix issues with CRD generation (in repos that don't have CRDs)

inteon:
Generates the CRDs using `CONTROLLER-GEN` in a temporary directory and only creates the `crds_dir` directory if that folder is non-empty.


erikgb (commented):
/lgtm
/approve

Thanks Tim! ❤️

PR #224, opened 2024-12-12
Bump kind version to v0.25.0

inteon:

PR #227, opened 2024-12-16
Pull preloaded images based on digest, not tag

inteon:
Fixes a long-standing issue where images were preloaded by tag instead of digest causing the digest check to fail when the upstream image changed.
Now, the images are pulled by digest, we modify the pulled tarball to have the desired image tag and we use the tarball that way.

Is a long-term fix for the issue we tried to fix here: https://github.com/cert-manager/makefile-modules/pull/224

### DIFF
--- a/modules/kind/kind-image-preload.mk
+++ b/modules/kind/kind-image-preload.mk
@@ -32,18 +32,33 @@ images_files := $(foreach image,$(images),$(subst :,+,$(image)))
 images_tar_dir := $(bin_dir)/downloaded/containers/$(HOST_ARCH)
 images_tars := $(images_files:%=$(images_tar_dir)/%.tar)
 
-# Download the images as tarballs. We must use the tag because the digest
-# will change after we docker import the image. The tag is the only way to
-# reference the image after it has been imported. Before downloading the
-# image, we check that the provided digest matches the digest of the image
-# that we are about to pull.
-$(images_tars): $(images_tar_dir)/%.tar: | $(NEEDS_CRANE)
-	@$(eval image=$(subst +,:,$*))
-	@$(eval image_without_digest=$(shell cut -d@ -f1 <<<"$(image)"))
-	@$(eval digest=$(subst $(image_without_digest)@,,$(image)))
-	@mkdir -p $(dir $@)
-	diff <(echo "$(digest)  -" | cut -d: -f2) <($(CRANE) manifest --platform=linux/$(HOST_ARCH) $(image_without_digest) | sha256sum)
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
Capturing discussion I remember from standup this morning:

- We might actually _want_ to break CI if a digest changes, to alert us to potential supply-chain attacks. It's not certain we want to avoid that.
- This PR essentially introduces a "cert-manager tag" which is pinned to the digest we first observed that tag pointing to. That means that someone running with a locally-pulled image might get different behavior to what we have, since we'll have a different digest and it'll be a bit hidden. It's not certain that we want that behavior.

I don't think either of the above are blockers and I don't have time to actually review this PR - just trying to capture the discussion here.

inteon:
> There's been a more recent kind release: https://github.com/kubernetes-sigs/kind/releases/tag/v0.26.0
> 
> I wonder if bumping to that is a short-term fix for the kind issues affecting CI?

See https://github.com/cert-manager/makefile-modules/pull/228

inteon:
> Capturing discussion I remember from standup this morning:
> 
> * We might actually _want_ to break CI if a digest changes, to alert us to potential supply-chain attacks. It's not certain we want to avoid that.
> * This PR essentially introduces a "cert-manager tag" which is pinned to the digest we first observed that tag pointing to. That means that someone running with a locally-pulled image might get different behavior to what we have, since we'll have a different digest and it'll be a bit hidden. It's not certain that we want that behavior.
> 
> I don't think either of the above are blockers and I don't have time to actually review this PR - just trying to capture the discussion here.

- I don't think it is very useful to break CI when the remote registry changes what digest a tag points to. We might want some notification/ alerts maybe? Also, if there was a supply-chain attack that got fixed, we would probably hear about it in another way. By pinning to the digest disregarding the upstream tag, we prevent upstream changes from breaking our flow and/or introducing any new supply-chain attacks.
- I don't think this will be a huge problem, the main offenders are kind's node image and busybox's image. For both, I think we are better off pinning the image digests and not breaking CI. This also ensures our tests are more reproducible (eg. when we want to patch an old version of one of our components and the upstream tag changed in the meantime).

hawksight:
Just adding here that I spent some time looking for other solutions by pulling various image in tarball and oci format. Basically what @inteon has done seems to be sensible.

I'd suggest that we trust kind images when they retag them, moving their version tags to new digests.

It seems kind does not support importing the OCI format images, only tarball. And when pulling a digest bases tarball... you don't get the actual tag in the manifest.json... even id you do <tag>@<digest>.

maelvls:
/lgtm

Seems like these changes have already been reviewed multiple times, but I haven't tested them manually myself.

PR #233, opened 2024-12-18
Fix self-upgrade failure due to invalid version returned by git describe in CI

inteon:
See https://github.com/cert-manager/trust-manager/actions/runs/12393729213 and https://github.com/cert-manager/trust-manager/actions/runs/12393952730 for the before and after resp.


inteon:
self-lgtm'ed since this is a small fix & I don't want this to fail during the holidays.

PR #241, opened 2025-01-28
Rename and add go build options and make them work for both oci-build and executable modules

inteon:
- added `go_$1_flags`
- added `oci_$1_linux_capabilities`
- renamed `cgo_enabled_$1` to `go_$1_cgo_enabled` (to match the naming pattern used for all other go-related variables)
- renamed `goexperiment_$1` to `go_$1_goexperiment` (to match the naming pattern used for all other go-related variables)
- renamed `oci_additional_layers_$1` to `oci_$1_additional_layers` (to match the naming pattern used for all other oci-related variables)
- `build_names` are no longer included in the `executable` module, explicitly use `exe_build_names` instead

This PR will fix the renamed variable for the trust-manager project (only OSS project that uses these variables):
https://github.com/cert-manager/trust-manager/pull/538

### DIFF
--- a/modules/executable/00_mod.mk
+++ b/modules/executable/00_mod.mk
@@ -1,0 +1,79 @@
+# Copyright 2023 The cert-manager Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Makes sense to me, happy to review follow ups if needed.

PR #239, opened 2025-01-24
Remove use of '-T' in cp (not supported on macOS)

SgtCoDFish:
This worked when tested on a closed source macOS project - haven't tested on Linux

PR #242, opened 2025-01-28
Fixing helm module CRD gen

SgtCoDFish:
This is a followup to #239, fixing the issue in a platform-independent way and cleaning up the old mistaken folder.

Once this merges and is propagated to all repos, the new `rm` added in this PR can be removed.

I tested this branch locally against a checkout of trust-manager and it had the expected effect - the new CRD was removed and nothing else changed.


wallrj (approved):
Thanks.

/approve
/lgtm

SgtCoDFish:
Testing:

Triggered self-upgrade in trust-manager: https://github.com/cert-manager/trust-manager/actions/runs/13015685538

That failed because of a dependabot PR which Tim merged today in the makefile-modules repo: https://github.com/cert-manager/makefile-modules/pull/240

Because that edited a github workflow file. So I'm going to abandon testing for now to avoid more yak shaving

SgtCoDFish:
Followed up with #243 - and will be completed with manual modules upgrades across all repos.

PR #243, opened 2025-01-29
Stop trying to rm CRDs dir

SgtCoDFish:
This was intended to help clean up the extra CRDs dir which was created by mistake in https://github.com/cert-manager/makefile-modules/pull/239.

However, a different makefile-modules PR (https://github.com/cert-manager/makefile-modules/pull/240) meant that a manual upgrade was required for downstream repos using makefile-modules anyway.

As such, this is just error-prone and liable to break ci (it should've been `rm -rf` anyway).

### DIFF
--- a/modules/helm/crds.mk
+++ b/modules/helm/crds.mk
@@ -47,11 +47,6 @@ generate-crds: | $(NEEDS_CONTROLLER-GEN) $(NEEDS_YQ)
 	$(eval crds_gen_temp := $(bin_dir)/scratch/crds)
 	$(eval directories := $(shell ls -d */ | grep -v -e 'make' $(shell git check-ignore -- * | sed 's/^/-e /')))
 
-	# TODO(@SgtCoDFish): This is a temporary fix for an accidentally created nested crds directory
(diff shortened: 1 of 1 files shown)


modules/helm/crds.mk
```diff
@@ -50,9 +50,8 @@ generate-crds: | $(NEEDS_CONTROLLER-GEN) $(NEEDS_YQ)
 	# TODO(@SgtCoDFish): This is a temporary fix for an accidentally created nested crds directory
 	# in the PR: https://github.com/cert-manager/makefile-modules/pull/239
 	# Once this has been run on every repo with the mistaken folder, this can be removed
-	rm -r $(crds_dir)/crds
+	rm -rf $(crds_dir)/crds
 
-	rm -rf $(crds_gen_temp)
```
inteon, line 55: This line was correct, I think, shouldn't we remove the line above and the comment?

SgtCoDFish: This is why we have code reviews! thank you!

SgtCoDFish: Fixed!

PR #244, opened 2025-01-29
Make the generate-govulncheck base dir configurable

inteon:
This will make it possible to modify the GH action and add authentication or replace the GH action with a GitLab pipeline.


SgtCoDFish (approved):
/lgtm
/approve

Makes sense!

PR #245, opened 2025-01-30
Bump openshift preflight to latest

SgtCoDFish:
The preflight tool's checks will fail if the latest version isn't being used


inteon (commented):
/approve
/lgtm

PR #249, opened 2025-02-04
Save the Helm chart digest to a file and use it to tag the non-v-prefixed Helm chart image

inteon:
This makes it easier to keep track of the image that was pushed during the release process.

Test run for approver-policy:
```
$ make release
...
RELEASE_OCI_MANAGER_IMAGE=127.0.0.1:8078/cert-manager-approver-policy
RELEASE_OCI_MANAGER_TAG=v0.17.0-35-gebfc292e527bbb-dirty
RELEASE_HELM_CHART_IMAGE=127.0.0.1:8078/charts/cert-manager-approver-policy
RELEASE_HELM_CHART_VERSION=v0.17.0-35-gebfc292e527bbb-dirty
Release complete!
$ crane ls 127.0.0.1:8078/charts/cert-manager-approver-policy
0.17.0-35-gebfc292e527bbb-dirty
v0.17.0-35-gebfc292e527bbb-dirty
```

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -46,6 +46,8 @@ helm_chart_image_registry := $(dir $(helm_chart_image_name))
 helm_chart_image_tag := $(helm_chart_version)
 helm_chart_sources := $(shell find $(helm_chart_source_dir) -maxdepth 1 -type f) $(shell find $(helm_chart_source_dir)/templates -type f)
 helm_chart_archive := $(bin_dir)/scratch/image/$(helm_chart_name)-$(helm_chart_version).tgz
+helm_digest_path := $(bin_dir)/scratch/image/$(helm_chart_name)-$(helm_chart_version).digests
+helm_digest = $(shell head -1 $(helm_digest_path) 2> /dev/null)
 
 $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_dir)/scratch/image
 	$(eval helm_chart_source_dir_versioned := $@.tmp)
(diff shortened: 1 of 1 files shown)


modules/helm/helm.mk
```diff
@@ -74,8 +76,11 @@ $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_d
 ## Will also create a non-v-prefixed tag for the OCI image.
 ## @category [shared] Publish
 helm-chart-oci-push: $(helm_chart_archive) | $(NEEDS_HELM) $(NEEDS_CRANE)
-	$(HELM) push "$(helm_chart_archive)" "oci://$(helm_chart_image_registry)"
-	$(CRANE) copy "$(helm_chart_image_name):$(helm_chart_image_tag)" "$(helm_chart_image_name):$(helm_chart_image_tag:v%=%)"
+	$(HELM) push "$(helm_chart_archive)" "oci://$(helm_chart_image_registry)" 2>&1 \
+		| grep -o "sha256:.\+" \
+		| tee $(helm_digest_path)
+
+	$(CRANE) copy "$(helm_chart_image_name)@$(call helm_digest)" "$(helm_chart_image_name):$(helm_chart_image_tag:v%=%)"
```
SgtCoDFish, line 83: suggestion (non-blocking): Not required to merge (it was there before this PR) but there are lot of magic incantations in various modules which would benefit from more comments. Documenting what `$(helm_chart_image_tag:v%=%)` does would help a lot for people less familiar with make / bash!

SgtCoDFish (approved):
/lgtm
/approve

This kind of change makes me a bit nervous - grepping output is so liable to break in the future because tools change their output all the time. I suppose if the grep does break, we should at least notice.

I don't want to get in the way so I'll approve, but I've said my bit about grepping output too!

modules/helm/helm.mk
inteon: The PR already merged, but I'll explain it here:
`$(helm_chart_image_tag:v%=%)` removes the `v` prefix from the value stored in `helm_chart_image_tag`.
See https://www.gnu.org/software/make/manual/html_node/Substitution-Refs.html for the manual on the syntax.

inteon:
> /lgtm /approve
> 
> This kind of change makes me a bit nervous - grepping output is so liable to break in the future because tools change their output all the time. I suppose if the grep does break, we should at least notice.
> 
> I don't want to get in the way so I'll approve, but I've said my bit about grepping output too!

I agree, grepping is non-ideal, but it seems to be the only way to securely obtain the digest atm.

PR #252, opened 2025-02-05
BUGFIX: load Helm digest from file after writing it

inteon:
All makefile templating is done before running the bash script, resulting in the digest file being read before it was actually written to.
Now, we read the digest in the bash script itself, fixing the issue.

Bug was introduced in https://github.com/cert-manager/makefile-modules/pull/249


inteon:
self-lgtm'ing to get this fix quickly

PR #260, opened 2025-03-07
Fix the verify-pod-security-standards target which was silently failing

wallrj:
Merging this might cause the verify-pod-security-standards target to start failing in projects when they auto-upgrade. I've only checked it with Firefly MR: !479

There were multiple problems:

1. The INSTALL_OPTIONS were not being used So for charts with required options the `helm template` command was failing, but the make target exited with 0. E.g. in firefly

```
$ make verify-pod-security-standards
...
Error: execution error at (firefly/templates/NOTES.txt:8:4): please supply the clientID for your Venafi TLS Protect Cloud service account, by setting the Helm value `deployment.venafiClientID: <CLIENT_ID>` or by supplying the command line option `--set deployment.venafiClientID=<CLIENT_ID>`.

Use --debug flag to render out invalid YAML

Applying 0 policy rule(s) to 0 resource(s)...

pass: 0, fail: 0, warn: 0, error: 0, skip: 0

$ echo $?
0
```

2. `kyverno apply` was not reading the policy from the process substitution. The command was exiting with 0 as follows:

```
$ make verify-pod-security-standards
...
Applying 0 policy rule(s) to 16 resource(s)...

pass: 0, fail: 0, warn: 0, error: 0, skip: 0

$ echo $?
0
```

3. There was no way to provide exceptions, for when the deployment contains expected violations.

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -64,7 +64,7 @@ $(helm_chart_archive): $(helm_chart_sources) | $(NEEDS_HELM) $(NEEDS_YQ) $(bin_d
 		echo "Chart name does not match the name in the helm_chart_name variable"; \
 		exit 1; \
 	fi
-	
+
 	$(YQ) '.annotations."artifacthub.io/prerelease" = "$(IS_PRERELEASE)"' \
 		--inplace $(helm_chart_source_dir_versioned)/Chart.yaml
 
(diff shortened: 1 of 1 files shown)


inteon (commented):
Thank you for looking into this, using `INSTALL_OPTIONS` seems like the perfect solution.

/approve
/lgtm

PR #261, opened 2025-03-17
Add example for a Kyverno policy exception

SgtCoDFish:
I struggled with this in a non-cert-manager PR (https://github.com/jetstack/jetstack-secure/pull/640) so I've added an example of what an exception looks like in this PR. Purely for documentation!


maelvls:
/lgtm
/approve

thank you 🙏

PR #266, opened 2025-04-07
Bump golang to 1.24

inteon:
Start using 1.24 instead of 1.23


maelvls:
We had a discussion on whether 1.24 would have any impact on any of the downstream projects (approver-policy, trust-manager, cert-manager, etc). Except for some unit tests failing in cmctl, nothing important is expected to break. So we have decided to go ahead an upgrade to Go 1.24.

/approve
/lgtm

PR #269, opened 2025-04-09
Change copy logic to fix macOS issues caused by 15.4

SgtCoDFish:
macOS 15.4 changed rsync to point to a version of [openrsync](https://derflounder.wordpress.com/2025/04/06/rsync-replaced-with-openrsync-on-macos-sequoia/), which in turn seems to use different methods for opening files. rsync is currently used to copy files in "make verify" targets.

The change to openrsync causes issues on macOS 15.4 locally for me when copying git objects:

> `rsync(35586): error: .git/objects/<num>/<sha>: openat: Permission denied`

By changing away from rsync and using find instead, I can reliably run "make verify" locally and remove our dependency on rsync.

### DIFF
--- a/modules/generate-verify/util/verify.sh
+++ b/modules/generate-verify/util/verify.sh
@@ -44,7 +44,17 @@ cleanup() {
 }
 trap "cleanup" EXIT SIGINT
 
-rsync -aEq "${projectdir}/." "${tmp}" --exclude "_bin/"
+# Why not just "cp" to the tmp dir?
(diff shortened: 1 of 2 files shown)


inteon:
This is working on macos, but seems to be failing on linux (see CI failure https://github.com/cert-manager/makefile-modules/actions/runs/14362278474/job/40322970066?pr=269).

inteon:
Thanks for fixing what macos broke.
/approve
/lgtm

PR #271, opened 2025-04-11
Use github.com/cert-manager/image-tool instead of in-tree go script

inteon:
Start using github.com/cert-manager/image-tool, which is based on the in-tree go scripts that we had in this repo.
By moving the project to a separate repo, we can further expand its functionality and use it across modules (eg. it is now also used in the kind module).

NOTE: I verified that this works for the "make test-smoke" target in approver-policy.


inteon:
WIP
/hold

SgtCoDFish (approved):
/lgtm
/approve

Wow, it really hadn't occurred to me that this was a thing we could do. Complete blind spot for me.

Thank you, this is awesome!

Issue #275, opened 2025-04-22
Limit govulncheck to cert-manager org


lentzi90:
Would it be ok to limit the [govulncheck](https://github.com/cert-manager/makefile-modules/blob/main/modules/go/base/.github/workflows/govulncheck.yaml) workflow to the cert-manager org (in the same way as self-upgrade was https://github.com/cert-manager/makefile-modules/pull/185)?
The vulnerability scanning *works* in forks but I think it is far better to not run them by default in every fork for multiple reasons:

1. Unless the fork `main` branch is kept up to date, the result is worthless
2. It wastes resources for all forks where the results are not wanted/checked
3. It may spam fork owners with notifications
4. If the fork is used for "internal" patches, the branch will likely be changed to something other than main anyway, so they could then change the org limitation as well.

PR #276, opened 2025-04-23
Limit govulncheck to cert-manager org

lentzi90:
Fixes: #275


inteon:
/approve
/lgtm

Thank you for the change, I think it makes sense to limit this GH action to cert-manager repos only.

PR #278, opened 2025-04-25
Don't make oci-layout folder path dependent on image tag

inteon:
This creates unnecessary additional complexity and leftover folders.


wallrj (approved):
I haven't tested this, but I guess you have.

My only worry is that this might cause the build to use stale files.
Adding the tag to the target directory ensures that the files have to be recreated everytime the git derived version changes...e.g. v1.0.0-commithash-numberofcommits

PR #280, opened 2025-04-25
Add kubeconform verification of helm charts

SgtCoDFish:
This catches the issue fixed by https://github.com/cert-manager/csi-driver-spiffe/pull/284 - so it's probably a useful thing to have in the Helm module, globally.

### DIFF
--- a/modules/helm/helm.mk
+++ b/modules/helm/helm.mk
@@ -178,3 +178,11 @@ verify-helm-lint: $(helm_chart_archive) | $(NEEDS_HELM)
 	$(HELM) lint $(helm_chart_archive)
 
 shared_verify_targets_dirty += verify-helm-lint
+
+.PHONY: verify-helm-kubeconform
+## Verify that the Helm chart passes a strict check using kubeconform
+## @category [shared] Generate/ Verify
+verify-helm-kubeconform: $(helm_chart_archive) | $(NEEDS_KUBECONFORM)
+	$(HELM) template kubeconform-template-do-not-use $< | $(KUBECONFORM) -strict
+
+shared_verify_targets_dirty += verify-helm-kubeconform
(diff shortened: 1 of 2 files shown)


modules/helm/helm.mk
```diff
@@ -178,3 +178,11 @@ verify-helm-lint: $(helm_chart_archive) | $(NEEDS_HELM)
 	$(HELM) lint $(helm_chart_archive)
 
 shared_verify_targets_dirty += verify-helm-lint
+
+.PHONY: verify-helm-kubeconform
+## Verify that the Helm chart passes a strict check using kubeconform
+## @category [shared] Generate/ Verify
+verify-helm-kubeconform: $(helm_chart_archive) | $(NEEDS_KUBECONFORM)
+	$(HELM) template kubeconform-template-do-not-use $< | $(KUBECONFORM) -strict
```
erikgb, line 186: Do not use? 😆

SgtCoDFish: Yeah that's the [NAME] of the chart:

>  helm template --help
>  helm template [NAME] [CHART]

So the rendered output will have the name kubeconform-template-do-not-use in it:

```yaml
# Source: cert-manager-csi-driver-spiffe/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  namespace: default
  labels:
    app.kubernetes.io/name: cert-manager-csi-driver-spiffe
    helm.sh/chart: cert-manager-csi-driver-spiffe-v0.0.0
    app.kubernetes.io/instance: kubeconform-template-do-not-use <---------------
    app.kubernetes.io/version: "v0.0.0"
    app.kubernetes.io/managed-by: Helm
  name: cert-manager-csi-driver-spiffe
```

SgtCoDFish: I'm trying to avoid anyone using that output for anything else - probably wouldn't happen, maybe more confusing than helpful?

erikgb: It was a bit confusing, but I understand it better now that you have explained it!

erikgb (commented):
I don't understand all of this, but I trust you to know what you are doing.

/lgtm
/approve

SgtCoDFish:
> I don't understand all of this, but I trust you to know what you are doing.

Haha we'll see if I do 😂

PR #281, opened 2025-04-26
Use INSTALL_OPTIONS before kubeconform and add datreeio to kubeconform sources

inteon:
Should resolves some of the failures we are seeing (like https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_trust-manager/611/pull-trust-manager-verify/1915945738441854976).

The INSTALL_OPTIONS will make sure we are able to render the Helm chart (eg. when some of the values are required values).
By adding the datreeio source, we should be able to include cert-manager resources in our Helm charts.

This check was added here: https://github.com/cert-manager/makefile-modules/pull/280


erikgb (commented):
/lgtm
/approve

PR #282, opened 2025-04-26
Fix typos in licenses module

inteon:
- replace `licences` with `licenses`.
- fix `warning: undefined variable 'license_layer_path_manager'`


inteon:
self-lgtm'ing since I want this small fix applied quickly, additional PRs can follow if necessary

PR #283, opened 2025-04-28
Save downloaded tools in user's home cache folder and share them across repos

inteon:
This should speed up local development across multiple repos.
Now, all downloads should be cached across repos.


maelvls:
I like this idea. I also have huge `_bin` folders across projects, and also have multiple checkouts of certain projects

Haven't tested but I don't think this is a big deal, we can revert it back later if anything goes wrong.

/lgtm

PR #285, opened 2025-04-28
kind image preload: simplify variables

inteon:
Make TAR, REPO, TAG and FULL variables directly available (no longer require depending on the `env-...` target).
I found this simplification while working on making cert-manager/cert-manager use more makefile modules.

### DIFF
--- a/modules/kind/kind-image-preload.mk
+++ b/modules/kind/kind-image-preload.mk
@@ -29,10 +29,9 @@ endif
 IMAGE_ARCH ?= $(HOST_ARCH)
 
 images := $(images_$(IMAGE_ARCH))
-images_files := $(foreach image,$(images),$(subst :,+,$(image)))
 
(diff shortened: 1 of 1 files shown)


maelvls (approved):
good change, haven't tried it manually.

/lgtm

modules/kind/kind-image-preload.mk
```diff
@@ -26,22 +26,21 @@ endif
 
 ##########################################
 
-images := $(images_$(HOST_ARCH))
-images_files := $(foreach image,$(images),$(subst :,+,$(image)))
+IMAGE_ARCH ?= $(HOST_ARCH)
```
maelvls, line 29: doesn't seem useful? I don't see a use-case where having `HOST_ARCH` and `IMAGE_ARCH` different would help (or even work)

PR #296, opened 2025-06-05
Make CRD condition expression configurable

inteon:
For the cert-manager/cert-manager repo, we want the expression to be `or .Values.crds.enabled .Values.installCRDs` instead of just `.Values.crds.enabled`.

See https://github.com/cert-manager/cert-manager/pull/7718 for the full PR.


erikgb (commented):
/lgtm
/approve
/hold

Could we deprecated the "non-standard" value for this in cert-manager and avoid this change? I am guessing there was a reason for renaming this value in newer projects (compared to cert-manager).

PR #300, opened 2025-06-17
Improve text in licenses template

SgtCoDFish:
As discussed in standup, adds a note for users.


inteon (commented):
/approve
/lgtm

Thanks @SgtCoDFish, this should cover everything (we now attribute the authors, we mention their licenses and point people to the correct version and license file contents).

PR #301, opened 2025-06-17
Fix location of licenses template

SgtCoDFish:
Fixes an issue in #300 when there are nested go.mod files within a repo.

I was a little wary of referring to `make/_shared` here but it's already referred to here: https://github.com/cert-manager/makefile-modules/blob/7085ba6f6b6c97fcd0062b692ac71513f6908c4c/modules/go/01_mod.mk#L33

### DIFF
--- a/modules/licenses/01_mod.mk
+++ b/modules/licenses/01_mod.mk
@@ -14,6 +14,9 @@
 
 ###################### Generate LICENSES files ######################
 
+# _module_dir is the directory containing this Makefile, used to retrieve the path of the licenses.tmpl file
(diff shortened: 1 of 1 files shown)


modules/licenses/01_mod.mk
```diff
@@ -14,6 +14,9 @@
 
 ###################### Generate LICENSES files ######################
 
+# _top is the directory containing the top-level Makefile
+_top = $(abspath $(dir $(firstword $(MAKEFILE_LIST))))
```
inteon, line 18: We use lastword in other places: https://github.com/cert-manager/makefile-modules/blob/main/modules/help/01_mod.mk#L16

PR #303, opened 2025-07-01
update commit to include latest bugfixes on fork

inteon:
Updated the fork to include new fixes:
https://github.com/google/go-licenses/compare/master...inteon:go-licenses:inteon_fork


maelvls:
Thanks for fixing these. Would be good to have a comment to remind me that this hash isn't an upstream one but rather a fork.

/lgtm
/approve

PR #304, opened 2025-07-04
Bump tools versions to their latest releases

erikgb:
Some of our versions were a bit outdated, but in general, this PR includes patch updates to tools. I bumped kustomize to v5, as the latest v4 release is from 2022. 😉


inteon (commented):
/approve
/lgtm

PR #307, opened 2025-07-11
Upgrade operator-sdk and add preflight darwin target

inteon:
Version `v1.40.0` had a replace statement, causing the installation to fail. We are now downloading the pre-built operator-sdk binaries from GitHub instead of building it using Go.

This commit also adds the darwin targets for the preflight tool, fixing a TODO comment.


wallrj (approved):
/approve
/lgtm

PR #311, opened 2025-07-22
Allow for excluding CRD template files

erikgb:
This adds a new feature to the Helm module, allowing exclusion of CRD(s) from getting included into the Helm chart. This feature is needed to avoid introducing the `ClusterBundle` API prematurely in trust-manger, ref. https://github.com/cert-manager/trust-manager/pull/662.

Generating the CRD will allow me to start working on validation, including tests, for the new API. But at the same time, avoiding the new CRD from being available to users.

### DIFF
--- a/modules/helm/crds.mk
+++ b/modules/helm/crds.mk
@@ -40,6 +40,15 @@ endif
 crds_dir ?= deploy/crds
 crds_dir_readme := $(dir $(lastword $(MAKEFILE_LIST)))/crds_dir.README.md
 crds_expression ?= .Values.crds.enabled
+crds_template_include_pattern := *.yaml
+# Space-separated list of basenames to exclude (e.g. foo.yaml *_test.yaml)
+crds_template_exclude_pattern ?=
+
+define filter-out-basenames
+  $(if $(strip $(2)), \
+    $(foreach f,$(1),$(if $(filter $(2),$(notdir $(f))),,$(f))), \
+    $(1))
+endef
 
 .PHONY: generate-crds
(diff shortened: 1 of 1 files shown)


modules/helm/crds.mk
```diff
@@ -40,6 +40,18 @@ endif
...
+crds_template_include_pattern := *.yaml
+# Space-separated list of basenames to exclude (e.g. foo.yaml *_test.yaml)
+crds_template_exclude_pattern ?=
+
+define filter-out-basenames
+  $(if $(strip $(2)), \
+    $(foreach f,$(1),$(if $(filter $(2),$(notdir $(f))),,$(f))), \
+    $(1))
+endef
+define crd-template-path
```
inteon, line 52: @erikgb I think we can keep this non-modifiable. There are no projects that don't use this pattern currently.

erikgb: I have updated it now. PTAL!

modules/helm/crds.mk
```diff
@@ -40,6 +40,15 @@ endif
 crds_dir ?= deploy/crds
 crds_dir_readme := $(dir $(lastword $(MAKEFILE_LIST)))/crds_dir.README.md
 crds_expression ?= .Values.crds.enabled
+crds_template_include_pattern := *.yaml
+# Space-separated list of basenames to exclude (e.g. foo.yaml *_test.yaml)
+crds_template_exclude_pattern ?=
+
+define filter-out-basenames
```
inteon, line 47: I don't understand how this works, but I trust you that it works.

PR #315, opened 2025-08-05
Add GOPRIVATE option for govulncheck

SgtCoDFish:
Full disclosure: this change is for a non-cert-manager repo. It should have no impact on any cert-manager stuff, but it'll make our go makefile-module a little easier to use elsewhere.

Simply adds the ability to specify GOPRIVATE for govulncheck.

I've tested that this allows a job to run with a private dependency: https://github.com/jetstack/jetstack-secure/actions/runs/16752151437

Example of what a run without this change looks like for that repo: https://github.com/jetstack/jetstack-secure/actions/runs/16751722851/job/47423444330


wallrj (approved):
/approve
/lgtm

PR #329, opened 2025-08-21
Move LICENSE file to boilerplate module

inteon:
This file fits the boilerplate module which checks for Apache-2.0 headers.
It allows users to use the repository-base module for non-Apache-2.0 repos.


inteon:
> If we merge this, won't it cause all the nightly makefile module updates to remove the LICENSE from open-source projects, because it is no longer in the repository base?
> 
> Have you tested it from another repo?
> 
> Merge if you're confident that it's not going to cause mayhem over the weekend.

All these projects also have the boilerplate module that now houses this LICENSE file.
It will only change projects that don't have the boilerplate module, which are non of the OSS projects.

PR #330, opened 2025-08-21
Make golangci-lint required

inteon:
This module is used by all repos, so we can just make it non-optional.


erikgb (commented):
Thanks, Tim! Looks like a good way forward!

/lgtm

PR #335, opened 2025-08-23
Replace Dependabot Go dependency upgrades with Renovate Bot

erikgb:
Dependabot isn't able to upgrade Go dependencies in projects with multiple (interconnected) Go modules. This has been discussed in multiple cert-manager bi-weekly meetings. It seems like Renovate provides a lot more flexibility than Dependabot, so it was decided to POC replacing Dependabot with Renovate in this area.

We have now tested Renovate in https://github.com/cert-manager/cert-manager and https://github.com/cert-manager/webhook-cert-lib, and I think we are ready for broader testing across our projects.

In this PR, we propose to replace the Dependabot Go dependency updates with Renovate. It is not strictly correct to include Renovate in the Dependabot (base) module, but that seems like the simplest way to achieve the goal here. And the naming can always be cleaned up later.

The Renovate config is identical to what we would have in cert-manager after merging https://github.com/cert-manager/cert-manager/pull/7999.


hjoshi123:
Ah thank you for explaining @ThatsMrTalbot. Sorry for the confusion here

PR #337, opened 2025-08-24
Add labels to Dependabot PRs

erikgb:
This should allow CI to run on Dependabot PRs automatically, so just a simple /lgtm and /approve is required to merge it. Example from cert-manager with the user action I would like to avoid: https://github.com/cert-manager/cert-manager/pull/8006


inteon (commented):
/approve
/lgtm

PR #345, opened 2025-08-26
ci(renovate): allow any postUpgradeTasks command

erikgb:
We were discussing running multiple postUpgradeTasks commands to fix a Renovate issue yesterday. As updating workflows from makefile-modules is a manual operation at present, I suggest allowing any command here. We control the Renovate config, so no need to be scared of the warnings here: https://docs.renovatebot.com/self-hosted-configuration/#allowedcommands

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -50,4 +50,4 @@ jobs:
           RENOVATE_ONBOARDING: "false"
           RENOVATE_PLATFORM: "github"
           LOG_LEVEL: "debug"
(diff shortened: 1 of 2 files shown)


hjoshi123:
Sounds reasonable.. it's a minor change with less implications since we control the pipeline
/approve

erikgb:
@inteon, I think this PR is a natural prerequisite to solve https://github.com/cert-manager/makefile-modules/pull/347#issuecomment-3227179805. PTAL!

PR #353, opened 2025-08-27
ci(renovate): adding more tools for renovate

hjoshi123:


modules/tools/00_mod.mk
```diff
@@ -170,14 +171,20 @@ tools += yamlfmt=v0.17.2
 tools += kubeconform=v0.7.0
 
 # https://pkg.go.dev/k8s.io/code-generator/cmd?tab=versions
-K8S_CODEGEN_VERSION := v0.33.3
```
erikgb, line 173: Have you looked at the potential consequences of deleting this variable? It's a breaking change.

hjoshi123: Oh right my bad.. totally spaced out on that.. although if we keep that variable how do we update the env through renovate?

modules/tools/00_mod.mk
```diff
@@ -69,6 +69,7 @@ tools += kubectl=v1.33.3
 # renovate: datasource=github-releases packageName=kubernetes-sigs/kind
 tools += kind=v0.29.0
 # https://www.vaultproject.io/downloads
+# renovate: datasource=github-releases packageName=hashicorp/vault
```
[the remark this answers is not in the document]
hjoshi123: We could change the way its installed below in the makefile.. I could make that change

modules/tools/00_mod.mk
erikgb: We probably need to modify the regex. 🤣 But if the variable isn't in use across the org, we can remove it. But that needs to be checked.

hjoshi123: yes I can check that.. I really dont want to touch the working regex 😛

modules/tools/00_mod.mk
```diff
@@ -69,6 +69,7 @@ tools += kubectl=v1.33.3
 # renovate: datasource=github-releases packageName=kubernetes-sigs/kind
 tools += kind=v0.29.0
 # https://www.vaultproject.io/downloads
+# renovate: datasource=docker packageName=hashicorp/vault
```
inteon, line 72: I think we should use datasource=go here too.

modules/tools/00_mod.mk
```diff
@@ -149,13 +162,16 @@ tools += cmctl=v2.3.0
 # https://pkg.go.dev/github.com/cert-manager/release/cmd/cmrel?tab=versions
 tools += cmrel=e3cbe5171488deda000145003e22567bdce622ea
 # https://pkg.go.dev/github.com/golangci/golangci-lint/v2/cmd/golangci-lint?tab=versions
+# renovate: datasource=go packageName=github.com/golangci/golangci-lint
```
erikgb, line 165:
```suggestion
# renovate: datasource=go packageName=github.com/golangci/golangci-lint/v2
```

modules/tools/00_mod.mk
```diff
@@ -149,13 +162,16 @@ tools += cmctl=v2.3.0
...
 tools += golangci-lint=v2.3.0
 # https://pkg.go.dev/golang.org/x/vuln?tab=versions
+# renovate: datasource=go packageName=golang.org/x/vuln
 tools += govulncheck=v1.1.4
 # https://github.com/operator-framework/operator-sdk/releases
 # renovate: datasource=github-releases packageName=operator-framework/operator-sdk
 tools += operator-sdk=v1.41.1
 # https://pkg.go.dev/github.com/cli/cli/v2?tab=versions
+# renovate: datasource=go packageName=github.com/cli/cli
```
erikgb, line 174:
```suggestion
# renovate: datasource=go packageName=github.com/cli/cli/v2
```

modules/tools/00_mod.mk
erikgb: I think this requires more careful thinking. Should we just postpone this change to a follow-up PR and get this merged?

PR #359, opened 2025-08-29
ci(renovate): take c-m Go deps out of misc group

erikgb:
This PR is motivated by https://github.com/cert-manager/google-cas-issuer/pull/307, where I think github.com/cert-manager/issuer-lib upgrades should be suggested in a separate PR. The same can be said about all our projects that depend on github.com/cert-manager/cert-manager. It has a lot of dependencies that often give issues when upgrading.


erikgb:
Not sure if this is a good idea after all.

/close

PR #387, opened 2025-09-01
fix(ci): changing glob patterns

hjoshi123:
We saw a bug https://github.com/cert-manager/istio-csr/pulls where the testing libs were not getting matched properly due to the glob patterns.


erikgb (commented):
Thanks a lot for fixing my Glob-mess, @hjoshi123! 🚀 

/lgtm
/approve

PR #399, opened 2025-09-04
ci(renovate): configure gitIgnoredAuthors to avoid PRs detected as edited

erikgb:
This could hopefully solve the issue shown here: (example): https://github.com/cert-manager/csi-driver/issues/451

image

Docs: https://docs.renovatebot.com/configuration-options/#gitignoredauthors

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,6 +2,10 @@
   $schema: 'https://docs.renovatebot.com/renovate-schema.json',
   enabled: true,
   gitAuthor: 'Renovate Bot <renovate-bot@users.noreply.github.com>',
+  gitIgnoredAuthors: [
+    'Renovate Bot <renovate-bot@users.noreply.github.com>',
+    'cert-manager-bot <cert-manager-bot@users.noreply.github.com>',
+  ],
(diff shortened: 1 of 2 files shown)


wallrj:
The root cause of the "detected as edited" symptom turned out to be the hard-coded `gitAuthor` not matching the identity the hosted Renovate app commits with — fix proposed in cert-manager/renovate-config#61, which has the full analysis.

One detail worth knowing: the `gitIgnoredAuthors` entries added here are `Name <email>` strings, but Renovate compares them against bare commit emails with no normalisation ([isBranchModified](https://github.com/renovatebot/renovate/blob/44.27.0/lib/util/git/index.ts#L1007-L1018), [setUserRepoConfig](https://github.com/renovatebot/renovate/blob/44.27.0/lib/util/git/index.ts#L404-L410)), so they never match — the `recreateWhen: 'always'` line is likely what actually worked around it. Once the `gitAuthor` fix lands, this repo's own `renovate.json5` can drop the hard-coded `gitAuthor`, these entries, and the `recreateWhen` TODO.

*with claude fable-5*

PR #403, opened 2025-09-04
Add command to delete obsolete Dependabot files

erikgb:
Since we are now migrating to Renovate.


modules/repository-base/01_mod.mk
```diff
@@ -34,6 +34,8 @@ else
 ## Generate base files in the repository
 ## @category [shared] Generate/ Verify
 generate-base:
+	# TODO(erikgb): Remove; just a temporary command to clean out Dependabot files
+	rm -f ./.github/dependabot.yaml $(repository_base_dependabot_dir)/.github/dependabot.yaml
```
ThatsMrTalbot, line 38: repository_base_dependabot_dir will be the one vendored inside make? This should automatically be cleaned up by klone I believe

PR #406, opened 2025-09-05
Disable Renovate for github.com/inteon/go-licenses dependency

erikgb:
This should make Renovate take the dependency out of https://github.com/cert-manager/makefile-modules/pull/401 and allow it to be merged.


inteon (commented):
/approve
/lgtm

PR #407, opened 2025-09-06
Enable Renovate vulnerability alerts

erikgb:
I want to see how this works, and eventually consider enabling it for our release branches.


erikgb:
It seems like Renovate requires read access to dependabot alerts to make this work, ref. https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts. And this is currently not a permission granted to the Octo STS App, ref. https://github.com/octo-sts/app/blob/3b34e02d9873acd86406d47341635ef5b898b149/README.md#octo-sts-github-permissions.

PR #455, opened 2025-10-29
Make OCI platforms configurable per OCI build target

inteon:
depends on https://github.com/cert-manager/makefile-modules/pull/457

Allows us to configure `oci_image1_platforms` and `oci_image2_platforms` separately.

### DIFF
--- a/modules/oci-build/01_mod.mk
+++ b/modules/oci-build/01_mod.mk
@@ -17,47 +17,54 @@ $(bin_dir)/scratch/image:
 
 .PHONY: $(oci_build_targets)
 ## Build the OCI image.
+## - oci-build-$(build_name) = build the oci directory (multi-arch)
+## - oci-build-$(build_name)__local = build the oci directory (local arch: linux/$(HOST_ARCH))
 ## @category [shared] Build
 $(oci_build_targets): oci-build-%: | $(NEEDS_KO) $(NEEDS_GO) $(NEEDS_YQ) $(NEEDS_IMAGE-TOOL) $(bin_dir)/scratch/image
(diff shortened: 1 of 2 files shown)


wallrj-cyberark:
> Allows us to configure oci_image1_platforms and oci_image2_platforms separately.

Why would we need to do that? cert-manager project has multiple images, but they're all built for the same set of platforms. Is it necessary for one of the csi projects?

modules/oci-build/01_mod.mk
```diff
@@ -17,47 +17,54 @@ $(bin_dir)/scratch/image:
...
 	KO_GO_PATH=$(GO) \
-	LDFLAGS="$(go_$*_ldflags)" \
-	$(KO) build $(go_$*_mod_dir)/$(go_$*_main_dir) \
-		--platform=$(oci_platforms) \
-		$(oci_$*_build_args) \
-		--oci-layout-path=$(oci_layout_path_$*) \
-		--sbom-dir=$(CURDIR)/$(oci_layout_path_$*).sbom \
+	LDFLAGS="$(go_$a_ldflags)" \
+	$(KO) build $(go_$a_mod_dir)/$(go_$a_main_dir) \
+		--platform=$(if $(is_local),linux/$(HOST_ARCH),$(oci_$a_platforms)) \
```
wallrj-cyberark: Don't mac users prefer to use `darwin` for local stuff? I can't remember. 
I think `--platform=""` has the same effect BTW. It causes `ko` to build only for your local platform.
I used that here: https://github.com/jetstack/jetstack-secure/blob/master/hack/ark/test-e2e.sh#L50, but admittedly, I don't think that script has ever been tried by a mac user.

PR #312, opened 2025-07-30
Simplify GCI golangci-lint config with localmodule

erikgb:
This PR simplifies the golangci-lint override config by taking advantage of a relatively new (last year) feature in the GCI linter: `localmodule`, avoiding the complex yq template syntax currently used.

PR #468, opened 2025-11-01
Simplify GCI golangci-lint config with localmodule

erikgb:
This is a (slightly modified) cherry-pick of https://github.com/cert-manager/makefile-modules/pull/312, which was later reverted in https://github.com/cert-manager/makefile-modules/pull/328. The missing piece was setting the `custom-order` setting to `true`, ref. https://golangci-lint.run/docs/formatters/configuration/#gci.

I have tested this again with golangci-lint 2.6.0, and it works now. And it seems like the `golangci-lint run --fix` proposed in https://github.com/cert-manager/makefile-modules/pull/467 requires us to use `localmodule` instead of `prefix` to do the right thing.

### DIFF
--- a/modules/go/.golangci.override.yaml
+++ b/modules/go/.golangci.override.yaml
@@ -70,10 +70,11 @@ formatters:
   enable: [ gci, gofmt ]
   settings:
     gci:
+      custom-order: true
       sections:
         - standard # Standard section: captures all standard packages.
         - default # Default section: contains all imports that could not be matched to another section type.
-        - prefix({{REPO-NAME}}) # Custom section: groups all imports with the specified Prefix.
+        - localmodule # Local module section: contains all local packages. This section is not present unless explicitly enabled.
         - blank # Blank section: contains all blank imports. This section is not present unless explicitly enabled.
         - dot # Dot section: contains all dot imports. This section is not present unless explicitly enabled.
   exclusions:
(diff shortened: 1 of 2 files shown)


inteon:
@erikgb this results in a LARGE diff in the cert-manager/cert-manager repo.

erikgb:
> > @erikgb this results in a LARGE diff in the cert-manager/cert-manager repo.
> 
> Ok, that's unfortunate. According to my testing, this should not be the case. Let me take a look.

I notice a diff, but only one diff in the main module of cert-manager, where `localmodule` now takes precedence over `dot`. The diff in the dedicated Go modules like cmd and test modules appears correct to me, as `localmodule` and `prefix(github.com/cert-manager/cert-manager)` don't always mean the same.

Should we revert this, @inteon? I would prefer keeping it, but I am also OK with a revert.

erikgb:
This was discussed in today's stand-up, where @SgtCoDFish, @ThatsMrTalbot, and @mladen-rusev-cyberark were present.

I want to know if we should roll forward or roll back this? Changing the import order to avoid the "templating" of the golangci-lint config is a one-time change, which can now be automated with the improved `fix-golangci-lint` makefile-modules target (ref. https://github.com/cert-manager/makefile-modules/pull/467).

We discussed this a bit, and none raised concern with a one-time change of the imports order/grouping. I will create a PR to "manually self-upgrade" in cert-manager. If this PR is approved and merged, we will move forward with this change. If not, I will ensure that we revert this change. CC: @inteon

PR #469, opened 2025-11-02
Disable modernize omitzero analyzer for now

erikgb:
This new analyzer generates a lot of new lint errors across our projects, and I think we need some time to discuss how to move forward with this.

See https://github.com/cert-manager/approver-policy/pull/739#issuecomment-3476839550 for context.


inteon:
@erikgb I think the linter is very useful and we can fix the linter without making breaking changes (eg. dropping omitempty instead of replacing it with omitzero). WDYT?

PR #594, opened 2026-04-17
move VENDORED_GO_VERSION to top of file

SgtCoDFish:
I open the tools module in my editor several times a week across different projects to answer the question "what Go version is this project using?"

This is a tiny quality-of-life fix which makes that question easier to answer, saving me from having to scroll / search and just putting the value at the top of the file.

If it's any help at all, here's what claude says about this very simple change:

```text
No, there's no reason to expect problems from this change.

  The uncommitted change in modules/tools/00_mod.mk simply moves the VENDORED_GO_VERSION variable definition from line ~226 to line ~35 (near the top of the file, after the directory
   definitions).

  This is just a code reorganization - the variable still has the same value (1.26.2) and will work exactly the same way. Moving it earlier in the file is actually logical since it's
   a fundamental configuration variable that might be referenced later in the makefile.

  The change is purely cosmetic and safe.```

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -30,6 +30,10 @@ endif
 export DOWNLOAD_DIR ?= $(default_shared_dir)/downloaded
 export GOVENDOR_DIR ?= $(default_shared_dir)/go_vendor
 
+# https://go.dev/dl/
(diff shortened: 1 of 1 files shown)


erikgb (commented):
/lgtm
/approve

Cheers! 🚀

PR #605, opened 2026-05-02
fix(helm): changing default test path for helm-unittest

hjoshi123:
helm-unittest has a default path of `tests/*.yaml` but that doesn't enable us to us different directories to structure our tests. This PR changes it to `tests/**/*.yaml`. Being a glob pattern it also matches `tests/*.yaml.` The same change will be done in cert-manager `ci.mk` to mimic makefile-modules.


erikgb (commented):
/lgtm
/approve

Makes sense!

PR #611, opened 2026-05-12
fix(tools): changing helm-unittest binary

hjoshi123:
Fixing the error from https://github.com/cert-manager/makefile-modules/pull/609. Helm unittest changed the binary after extraction to reflect the os and the arch. This PR addresses that change too.

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -117,7 +117,7 @@ tools += istioctl=1.29.2
 tools += controller-gen=v0.21.0
 # https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=versions
 # renovate: datasource=go packageName=golang.org/x/tools
-tools += goimports=v0.45.0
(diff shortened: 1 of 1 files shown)


modules/tools/00_mod.mk
```diff
@@ -117,7 +117,7 @@ tools += istioctl=1.29.2
 tools += controller-gen=v0.21.0
 # https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=versions
 # renovate: datasource=go packageName=golang.org/x/tools
-tools += goimports=v0.44.0
+tools += goimports=v0.45.0
```
erikgb: Could we revert unrelated changes like this one from the PR?

erikgb: @hjoshi123, I added a commit to fix this.

erikgb (commented):
/lgtm
/approve

Thanks! 🚀

PR #625, opened 2026-05-29
Add cache validation

inteon:
Add verify-cache command that validates the cache, preventing cache poisoning.

### DIFF
--- a/modules/tools/00_mod.mk
+++ b/modules/tools/00_mod.mk
@@ -39,6 +39,7 @@ $(bin_dir)/tools $(DOWNLOAD_DIR)/tools:
 
 checkhash_script := $(dir $(lastword $(MAKEFILE_LIST)))/util/checkhash.sh
 lock_script := $(dir $(lastword $(MAKEFILE_LIST)))/util/lock.sh
+verify_cache_script := $(dir $(lastword $(MAKEFILE_LIST)))/util/verify_cache.sh
 
 # $outfile is a variable in the lock script
 # Escape the dollar sign so it's passed literally to the shell script, not expanded by make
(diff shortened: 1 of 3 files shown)


modules/tools/00_mod.mk
```diff
@@ -785,3 +786,17 @@ go-tools: $(go_tool_names:%=$(bin_dir)/tools/%)
...
+
+.PHONY: verify-cache
+## Verify the integrity of $(DOWNLOAD_DIR) after restoring from an untrusted
+## cache. Checks SHA-256 of downloaded binaries, removes files with mismatched
+## hashes and unknown files, and verifies the Go module cache via go.sum. Exits
+## non-zero if any corruption is detected. Intended to be run after cache
+## restore in CI environments where the cache write-side is not fully trusted
+## (e.g. node-local hostPath shared with low-trust presubmit jobs).
+## @category [shared] Tools
+verify-cache:
```
maelvls, line 798: Does it mean we need to run `make verify-cache` in every single Prow job? I imagine that's not a big downside, but I wish it was a bit more automatic

inteon: Yes, I'll update the prowjob definitions to first call `make verify-cache`.

maelvls:
at first, I didn't believe it as I thought that the inode modification timestamp mismatch would prevent the attack. And I thought that the report was wrong as it mentioned poisonning `_bin/downloaded/kind` which doesn't actually exists...

I realized that I was wrong. 

```console
$ echo $'#!/bin/bash\necho I GOT YOU>&2'> ~/.cache/makefile-modules/downloaded/tools/kind@v0.31.0_darwin_arm64 && chmod +x ~/.cache/makefile-modules/downloaded/tools/kind@v0.31.0_darwin_arm64
$ make _bin/tools/kind
$ _bin/tools/kind
I GOT YOU
```

I'm surprised that we aren't verifying the hash when creating the symlink. Right now, we verify after downloading and then never verify again.

And since some artifacts are large tarballs, we would need check their tarballs' hash and untar them just before creating the symlink.

My recommendation would be exactly this: instead of uncompressing + checking hash on download, I'd uncompress + check hash upon creating the symlink. This would 100% mitigate the issue IMO.

Quick proof of concept for verifying hashes at runtime instead of at the time of download: https://github.com/cert-manager/cert-manager/pull/8833

wallrj:
Thanks @inteon — this has grown into the right shape: hash-check every restored file against the reviewed SHA-256 in `00_mod.mk`, purge anything unknown or mismatched, and `go mod verify` the module cache. The reviewed hash is the trust anchor, so a poisoned file simply fails to match and is removed — no signing or fork-scoping needed.

Two thoughts:

1. **It's a step you have to remember.** `verify-cache` only protects a job that calls it, after restore and *before* any tool runs. Nothing enforces that ordering, so a job that forgets is silently unprotected. @maelvls' cert-manager/cert-manager#8833 verifies at symlink-creation time, which has the opposite trade-off: automatic, but only fires for `$(NEEDS_*)` targets and is bypassed by anything that runs `_bin/tools/<x>` off `PATH` directly. The strongest version is content-addressing: key the cache directory by the reviewed hash, so a poisoned file isn't found under its expected key and verification collapses into a lookup — no separate step, no `PATH` bypass. Short of that, at least wire `verify-cache` in as a documented, mandatory first step after every cache restore.

2. **`go mod verify` is belt-and-braces.** `GOMODCACHE` is already verified against `go.sum`/GOSUMDB on every use, so a poisoned module cache is caught at build time regardless. No harm keeping it, but it's not where the exposure is — the tool cache, verified only at download until this PR, is.

Net: I'd like the tool-cache half of this to land here in makefile-modules, folding in cert-manager/cert-manager#8833's "verify at point of use, not just at download" insight rather than running it as a separate command. Happy to help push it over the line.

*with claude fable-5*

PR #636, opened 2026-06-06
Replace self-upgrade workflow with Renovate JSONata manager

erikgb:
The JSONata experiment added by @ThatsMrTalbot in https://github.com/cert-manager/helm-tool appears to be working now. This is how the self-upgrade PRs might look like after this change: https://github.com/cert-manager/helm-tool/pull/272.

We currently schedule all self-upgrade jobs at the same cron schedule, which makes the Prow jobs fail regularly - probably because of overload in the Prow cluster. With this change, Renovate will ensure the changes are distributed in a more randomly timed fashion.

We probably want to automerge the makefile-module upgrades, as we currently do, but I think we can do it in a follow-up PR. But I am a bit worried about noise, since Renovate will kick in more often than once a day - if not configured to do something else.


renovate-config.json5
```diff
@@ -3,7 +3,36 @@
   extends: [
     'github>cert-manager/renovate-config:default.json5',
   ],
+  "customManagers": [
+    {
+      "customType": "jsonata",
+      "fileFormat": "yaml",
+      "managerFilePatterns": ["klone.yaml"],
+      "matchStrings": [
+        'targets.*.{\
```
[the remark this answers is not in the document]
erikgb: @ThatsMrTalbot, you are our JSONata expert. WDYT?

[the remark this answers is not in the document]
ThatsMrTalbot: Fine as is:
https://try.jsonata.org/Ee2PoJ7HJ

renovate-config.json5
```diff
@@ -3,7 +3,36 @@
...
   packageRules: [
+    {
+      "groupName": 'Makefile Modules',
+      "matchManagers": ["custom.jsonata"],
+      "matchFileNames": ["klone.yaml"],
+      "postUpgradeTasks": {
+        "commands": [
+          "make vendor-go",
+          "make generate-klone",
+          "make generate"
```
SgtCoDFish, line 32: note: I think there's something interesting here I'd not thought enough about before.

`vendor-go` is in the tools module and `generate-klone` is in the klone module. `generate` is a target built from several modules. In this PR we're changing the repository-base module.

I think the implication is that to use this self upgrade flow (maybe to use makefile modules _at all_?) you need _at least_ the repository-base, tools and klone modules.

Makes me think that our separation of modules isn't quite right - we have really important cross dependencies across these three core modules. I wonder if it would make sense to condense these modules down into a single `base` module, conceptually.

(Not a blocker or anything because this is only for our consumption and we use those modules everywhere anyway - I'm just thinking out loud)

SgtCoDFish: There is actually a meaningful implication of this - the `make-self-upgrade` GH action used the `setup-go` action to provision go, so the repository-base + klone modules were all that were required for self-upgrades. Now in this PR we add a (hard) dependency on the tools module because we're using `make vendor-go` instead.

I don't think that's a blocker either, but it's worth calling out.

erikgb: Good point! Do we really need to have an exact Go version when running `make generate-klone` and `make generate`? Or would it be sufficient to have a Go version newer than version X?

erikgb: I had another look, and while I agree on your problem statement about module separation, I think this could be fixed in a follow-up PR I'll vote in favour of moving the vendor-go and unvendor-go targets to the repository-base module. And also merge the klone module into the repository-base module. WDYT, @inteon?

But in the context of this PR, I propose to keep it as proposed. I just tested this PR in the website project, which is our project with less Go content, and it works.

renovate-config.json5
```diff
@@ -3,7 +3,36 @@
   extends: [
     'github>cert-manager/renovate-config:default.json5',
   ],
+  "customManagers": [
```
SgtCoDFish, line 6: suggestion(claude): Found by AI, edited by me. I've not dug into checking if this is legit, so I'm not going to consider it a blocker or anything - but if you think it's legit + worth fixing, I'm happy to re-review next week!

## Missing executionMode: 'branch'

The new packageRules entry for custom.jsonata is missing `executionMode: 'branch'`, which every other rule with postUpgradeTasks in the repo uses.

Without it, Renovate defaults to per-update mode, meaning the three `make` commands run once per updated module entry. For a typical klone.yaml with several modules all pinned to the same hash those commands would run 7 times on a single branch. With `executionMode: 'branch'`, they run once for the whole branch.

SgtCoDFish (approved):
/lgtm
/approve

/hold

Hold because of the potential finding, but if you think it's bogus just unhold and merge.

renovate-config.json5
erikgb: Ok, this is interesting. I am going to POC this in the helm-tool project.

erikgb: https://github.com/cert-manager/helm-tool/pull/279

erikgb: Ok, this works, and I have updated the PR to include `executionMode: 'branch'`.

SgtCoDFish (approved):
/lgtm
/approve

Thank you for the bonus fix!

